PR #13, opened 2024-11-20
feat(backend): GET /applications

christophenne:


cmd/main.go
```diff
@@ -4,29 +4,41 @@ import (
...
+	router.Mount("/api", server.ApiRouter())
 	router.Handle("/*", StaticFileHandler())
 
+	go func() {
+		sigint := make(chan os.Signal, 1)
+		signal.Notify(sigint, syscall.SIGTERM, syscall.SIGINT)
+		<-sigint
+		_ = server.Shutdown()
+		fmt.Println("ok bye")
+		os.Exit(1)
```
kosmoz:
```suggestion
		os.Exit(0)
```

docker-compose.yaml
```diff
@@ -0,0 +1,16 @@
+version: '3.8'
```
kosmoz: I get this warning: `the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion`
```suggestion
```

internal/handlers/applications.go
```diff
@@ -0,0 +1,60 @@
...
+		r.Get("/", getApplication)
+	})
+}
+
+func getApplications(w http.ResponseWriter, r *http.Request) {
+	if applications, err := db.GetApplications(r.Context()); err != nil {
+		internalctx.GetLoggerOrPanic(r.Context()).Error("failed to get applications", zap.Error(err))
+		w.WriteHeader(http.StatusInternalServerError)
+	} else {
+		err := json.NewEncoder(w).Encode(applications)
```
kosmoz, line 28: the response has a wrong content type header but I think we can worry about that later

kosmoz: https://github.com/go-chi/render would help with this too

PR #15, opened 2024-11-21
feat: add create application endpoint

kosmoz:


internal/handlers/applications.go
```diff
@@ -2,24 +2,41 @@ package handlers
...
+	r.Post("/", createApplication)
 	r.Route("/{applicationId}", func(r chi.Router) {
 		r.Use(applicationMiddleware)
 		r.Get("/", getApplication)
 	})
 }
 
+func createApplication(w http.ResponseWriter, r *http.Request) {
+	var application types.Application
+	defer r.Body.Close()
```
christophenne: from: https://pkg.go.dev/net/http#Request
`The Server will close the request body. The ServeHTTP .Handler does not need to.`

internal/handlers/applications.go
```diff
@@ -2,24 +2,40 @@ package handlers
...
 }
 
+func createApplication(w http.ResponseWriter, r *http.Request) {
+	var application types.Application
+	if err := json.NewDecoder(r.Body).Decode(&application); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	} else if err = db.SaveApplication(r.Context(), &application); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		_, _ = fmt.Fprintln(w, err)
```
christophenne: I think this should be logged with the logger instead

internal/server/router.go
```diff
@@ -2,16 +2,25 @@ package server
...
 
 func ApiRouter() chi.Router {
 	// TODO for all (most) routes auth middleware
 	router := chi.NewRouter()
-	router.Use(loggerCtxMiddleware, dbCtxMiddleware)
+	router.Use(
+		middleware.RequestID,
+		loggerCtxMiddleware,
+		loggingMiddleware,
+		middleware.Recoverer,
```
christophenne: `middleware.Recoverer` is already "globally" added in `cmd/main.go`

PR #19, opened 2024-11-22
feat(cloud-ui): add color scheme switcher

pmig:


frontend/cloud-ui/src/app/components/side-bar/side-bar.component.ts
```diff
@@ -1,4 +1,4 @@
-import {Component} from '@angular/core';
```
kosmoz: unused imports

frontend/cloud-ui/src/app/components/color-scheme-switcher/color-scheme-switcher.component.html
```diff
@@ -0,0 +1,30 @@
+<button
+  id="theme-toggle"
+  data-tooltip-target="tooltip-toggle"
+  type="button"
+  (click)="switchColorScheme()"
+  class="text-gray-500 inline-flex items-center justify-center dark:text-gray-400 hover:bg-gray-100 w-10 h-10 dark:hover:bg-gray-700 focus:outline-none focus:ring-4 focus:ring-gray-200 dark:focus:ring-gray-700 rounded-lg text-sm p-2.5">
+  <svg
+    *ngIf="colorSchemeSignal() === ''"
```
kosmoz, line 8: For the future, let's use the new control flow sytnax in templates. For now it's OK, we can change it later.

Reference: https://angular.dev/guide/templates/control-flow

PR #27, opened 2024-11-26
feat: application versions endpoints

christophenne:


internal/handlers/applications.go
```diff
@@ -21,6 +24,18 @@ func ApplicationsRouter(r chi.Router) {
 		r.Use(applicationMiddleware)
 		r.Get("/", getApplication)
 		r.Put("/", updateApplication)
+		r.Route("/versions", func(r chi.Router) {
+			// note that it would not be necessary to use the applicationMiddleware for the versions endpoints
+			// it loads the application from the db including all versions, but I guess for now this is easier
+			// when performance becomes more important, we should avoid this and do the request on the database layer
```
kosmoz, line 30: I agree that it most likely won't matter for a while with regards to performance, but I think you could easily fix this by grouping the handlers that need the middleware together with the middleware. something like this: 

```go
r.With(applicationMiddleware).Group(func(r chi.Router) {
	r.Get("/", getApplication)
	r.Put("/", updateApplication)
})
r.Route("/versions", func(r chi.Router) {
	// ...
})
```

christophenne: Oh that's good to know. 

But how the handlers are now implemented, the middleware is necessary because I am using the `application`'s `Versions` field. Do you think I should change it and move the logic to the `db` layer?

kosmoz: no I think it's fine

internal/db/applications.go
```diff
@@ -38,19 +27,55 @@ func UpdateApplication(ctx context.Context, application *types.Application) erro
...
+func GetApplications(ctx context.Context) ([]types.Application, error) {
+	db := internalctx.GetDbOrPanic(ctx)
+	if rows, err := db.Query(ctx, `
+			SELECT a.id,
+			       a.created_at,
+			       a.name,
+			       a.type,
+			       CASE WHEN count(av.id) > 0
+			           THEN array_agg(row(av.id, av.created_at, av.name))
+						 END as versions
```
kosmoz: I think it would be nicer to return an empty array if there are no versions. currently, the get versions endpoint just returns `null`

internal/handlers/applications.go
```diff
@@ -88,6 +103,126 @@ func getApplication(w http.ResponseWriter, r *http.Request) {
...
+	} else if err = json.NewEncoder(w).Encode(applicationVersion); err != nil {
+		log.Error("failed to encode json", zap.Error(err))
+	}
+}
+
+func getApplicationVersionComposeFile(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := internalctx.GetLoggerOrPanic(ctx)
+	application := internalctx.GetApplicationOrPanic(ctx)
+	applicationVersionId := chi.URLParam(r, "applicationVersionId")
```
kosmoz: btw, you can also use r.PathValue instead, which decouples the handler from the routing lib

internal/handlers/applications.go
```diff
@@ -88,6 +103,126 @@ func getApplication(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	if file, _, err := r.FormFile("file"); err != nil {
+		if !errors.Is(err, http.ErrMissingFile) {
+			log.Error("failed to get file from upload", zap.Error(err))
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
```
kosmoz, line 147: so the compose file is optional?

christophenne: That's what I assumed, since we will have different types in the future, and so far its also nullable in the database.

kosmoz: OK just wanted to make sure this was intended

internal/types/data.go
```diff
@@ -1,9 +1,21 @@
...
 	Base
-	Name string         `db:"name" json:"name"`
-	Type DeploymentType `db:"type" json:"type"`
+	Name     string               `db:"name" json:"name"`
+	Type     DeploymentType       `db:"type" json:"type"`
+	Versions []ApplicationVersion `db:"versions" json:"versions"`
+}
+
+type ApplicationVersion struct {
+	// unfortunately Base nested type doesn't work when ApplicationVersion is a nested row in an SQL query
```
christophenne, line 13: I'd leave it at that for now.

kosmoz: fine my me

PR #34, opened 2024-11-26
feat(backend): add deployment target api

kosmoz:


internal/context/data.go
```diff
@@ -20,3 +20,17 @@ func WithApplication(ctx context.Context, application *types.Application) contex
 	ctx = context.WithValue(ctx, ctxKeyApplication, application)
 	return ctx
 }
+
+func WithDeploymentTarget(ctx context.Context, dt *types.DeploymentTarget) context.Context {
+	return context.WithValue(ctx, ctxKeyDeploymentTarget, dt)
+}
+
+func GetDeploymentTarget(ctx context.Context) *types.DeploymentTarget {
```
christophenne, line 28: It's a nicer name, I will remove the `OrPanic` suffix from the other methods in my PR so that they are consistent.

kosmoz: Yeah, I thought it was a little superflous

PR #36, opened 2024-11-26
feat(ui): add deployment targets ui

kosmoz:


frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.ts
```diff
@@ -0,0 +1,69 @@
...
+import {DeploymentTargetsService} from '../services/deployment-targets.service';
+import {AsyncPipe, DatePipe, NgOptimizedImage} from '@angular/common';
+import {FaIconComponent} from '@fortawesome/angular-fontawesome';
+import {FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators} from '@angular/forms';
+import {faCaretDown, faMagnifyingGlass, faPen, faPlus, faTrash, faXmark} from '@fortawesome/free-solid-svg-icons';
+import {DeploymentTarget} from '../types/deployment-target';
+import {lastValueFrom} from 'rxjs';
+
+@Component({
+  selector: 'app-deployment-targets',
```
christophenne, line 11: IntelliJ gives me the warning that this component is not declared in any module. I think in that case the component should be `standalone: true`

frontend/cloud-ui/src/app/services/deployment-targets.service.ts
```diff
@@ -0,0 +1,29 @@
...
+import {inject, Injectable} from '@angular/core';
+import {CrudService} from './interfaces';
+import {DeploymentTarget} from '../types/deployment-target';
+import {Observable, tap} from 'rxjs';
+import {DefaultReactiveList} from './cache';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DeploymentTargetsService implements CrudService<DeploymentTarget> {
```
christophenne, line 11: Not sure about the structure yet, I put the `ApplicationsService` into the `applications` feature directory. But we can also put all services in a `services` folder.

frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.ts
kosmoz: this is the default since v19, it's no longer needed

frontend/cloud-ui/src/app/services/deployment-targets.service.ts
kosmoz: Yes, I saw that but I didn't want to move it in case you wanted to make some changes too. 

IMO it's a little neater to have at least all CRUD type services in one place, but I don't really mind either way.

We should definitely decide one way or the other though :smile:

PR #44, opened 2024-11-27
feat: add limit to body size and file type for compose file

kosmoz:


internal/handlers/applications.go
```diff
@@ -144,18 +144,31 @@ func createApplicationVersion(w http.ResponseWriter, r *http.Request) {
...
 	} else {
-		applicationVersion.ComposeFileData = &data
+		log.Sugar().Debugf("got file %v with type %v and size %v", head.Filename, head.Header, head.Size)
+		// max file size is 100KiB
+		if head.Size > 102400 {
+			log.Debug("large body was rejected")
+			w.WriteHeader(http.StatusBadRequest)
+			fmt.Fprintln(w, "file too large (max 100 KiB)")
+			return
+		} else if head.Header.Get("Content-Type") != "application/yaml" {
```
christophenne: Here and at the upload input above we should also allow `application/x-yaml`, `text/yaml` and `text/x-yaml` like in https://github.com/glasskube/glasskube/blob/18a22272a7341fa07947d81663be11b66e5d344d/internal/contenttype/mediatype.go#L11

PR #46, opened 2024-11-28
feat: show deployment target status

christophenne:


frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.html
```diff
@@ -97,10 +98,28 @@
             -->
             <th scope="row" class="px-4 py-3 font-medium text-gray-900 whitespace-nowrap dark:text-white">
               <div class="flex items-center mr-3">
-                <img ngSrc="/docker.png" alt="Docker" class="h-8 w-auto mr-3 rounded" height="199" width="199" />
+                <div class="relative mr-3">
+                  <img ngSrc="/docker.png" alt="Docker" class="h-8 w-auto rounded" height="199" width="199" />
+                  @if (dt.currentStatus) {
```
christophenne, line 103: Probably good enough for now? Long term I think we need some enum type in the `DeploymentTargetStatus` that says healthy/not healthy, or maybe a bool suffices

PR #54, opened 2024-11-28
feat: deployments

pmig:


frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.ts
```diff
@@ -35,16 +48,26 @@ export class DeploymentTargetsComponent {
...
   private readonly overlay = inject(OverlayService);
   private readonly viewContainerRef = inject(ViewContainerRef);
 
   private readonly deploymentTargets = inject(DeploymentTargetsService);
-  readonly deploymentTargets$ = this.deploymentTargets.list();
+  readonly deploymentTargets$ = this.deploymentTargets.list().pipe(
+    map((dts) =>
+      dts.map((dt) => {
+        if (dt.id) {
+          dt.latestDeployment = this.deploymentTargets.latestDeploymentFor(dt.id);
```
christophenne: Should be moved to the DB as a join. Maybe with an optional `includeLatestDeployment` query parameter at the `/deployment-targets` endpoint. (after video)

internal/handlers/deployments.go
```diff
@@ -0,0 +1,80 @@
...
+		if _, err = fmt.Fprintln(w, err); err != nil {
+			log.Error("failed to write error to response", zap.Error(err))
+		}
+	} else if err = json.NewEncoder(w).Encode(deployment); err != nil {
+		log.Error("failed to encode json", zap.Error(err))
+	}
+}
+
+func getDeployments(w http.ResponseWriter, r *http.Request) {
+	deploymentTargetId := r.URL.Query().Get("deploymentTargetId")
```
christophenne, line 45: In the underlying `db` method the `deploymentTargetId` is required (otherwise SQL syntax error because an empty string is not a valid `uuid`), so one cannot call `/api/deployments` right now. 

Can also be done as soon as its needed, but I think getting all deployments is probably also a fair use case?

frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.html
```diff
@@ -299,3 +317,86 @@ <h3 class="text-xl font-semibold text-gray-900 dark:text-white">Setup Instructio
...
+        <form class="p-4 md:p-5" [formGroup]="deployForm" (ngSubmit)="saveDeployment()">
+          <div class="grid gap-4 mb-4 grid-cols-2">
+            <div class="col-span-2 sm:col-span-1">
+              <label for="applicationId" class="block mb-2 text-sm font-medium text-gray-900 dark:text-white"
+                >Application</label
+              >
+              <select
+                id="applicationId"
+                formControlName="applicationId"
+                (change)="onApplicationIdChange($event)"
```
christophenne: Not that it matters, but the way I'm used to do something like this is: 

```
this.deployForm.controls.applicationId.valueChanges.subscribe(appId => {
      // your logic
    });
```

You have to take care of the subscription though (if that is still valid for modern modern Angular).

frontend/cloud-ui/src/app/services/deployment.service.ts
```diff
@@ -0,0 +1,24 @@
...
+})
+export class DeploymentService {
+  private readonly baseUrl = '/api/deployments';
+  private readonly httpClient = inject(HttpClient);
+
+  create(request: Deployment): Observable<Deployment> {
+    return this.httpClient.post<Deployment>(this.baseUrl, request);
+  }
+
+  listForDeploymentTarget(deploymentTargetId: string | undefined): Observable<Deployment[]> {
```
christophenne: seems to be unused

frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.html
```diff
@@ -299,3 +317,86 @@ <h3 class="text-xl font-semibold text-gray-900 dark:text-white">Setup Instructio
...
+            </div>
+            <div class="col-span-2 sm:col-span-1">
+              <label for="applicationVersion" class="block mb-2 text-sm font-medium text-gray-900 dark:text-white"
+                >Version</label
+              >
+              <select
+                id="applicationVersion"
+                formControlName="applicationVersionId"
+                class="bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-primary-500 focus:border-primary-500 block w-full p-2.5 dark:bg-gray-600 dark:border-gray-500 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500"
+                [disabled]="!validApplicationSelected()">
```
christophenne:
```suggestion
                [disabled]="deployForm.controls.applicationId.invalid">
```

frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.ts
```diff
@@ -114,4 +137,58 @@ export class DeploymentTargetsComponent {
...
+  onApplicationIdChange($event: any) {
+    this.updatedSelectedApplication($event.target.value);
+  }
+
+  async updatedSelectedApplication(applicationId: string) {
+    let applications = await firstValueFrom(this.applications$);
+    this.selectedApplication = applications.find((a) => a.id === applicationId) || null;
+  }
+
+  validApplicationSelected() {
```
christophenne: removable when the above suggestion is used

internal/types/data.go
```diff
@@ -15,7 +15,20 @@ type ApplicationVersion struct {
...
+	ApplicationId   string    `db:"application_id" json:"applicationId"`
+}
+
+type Deployment struct {
+	Base
+	DeploymentTargetId   string `db:"deployment_target_id" json:"deploymentTargetId"`
+	ApplicationVersionId string `db:"application_version_id" json:"applicationVersionId"`
+}
+
+type DeploymentWithData struct {
```
christophenne, line 27: `DeployedApplication` / `DeploymentWithApplication`?

christophenne: maybe even all of these additional fields fit in the `Deployment` and can be left empty, but I don't know. (after video)

frontend/cloud-ui/src/app/types/deployment-target.ts
```diff
@@ -1,10 +1,13 @@
 import {BaseModel, Named} from './base';
 import {Geolocation} from './geolocation';
+import {DeploymentWithData} from './deployment';
+import {Observable} from 'rxjs';
 
 export interface DeploymentTarget extends BaseModel, Named {
   name: string;
   type: string;
   geolocation?: Geolocation;
+  latestDeployment?: Observable<DeploymentWithData>;
```
kosmoz: I think this style of relation mapping could be interesting but it should be implemented in a more deliberate way. Definitely, I believe everything in the DTO interface should be in the API response and if any additional behavior is implemented, it should be done in a class implementing the interface. However, at that point, it could be argued that a less "OO" approach of just using a function (albeit on a service) would be more ergonomic, since you don't have to keep both the DTO interface and class up-to-date. 

For now, I would suggest simply calling the service function from the template. Angular's change detection should be able to correctly keep the method calls per instance in line (I think).

frontend/cloud-ui/src/util/dates.ts
```diff
@@ -6,7 +6,7 @@ export function isOlderThan(date: dayjs.ConfigType, duration: Duration): boolean
   return dayjs.duration(Math.abs(dayjs(date).diff(dayjs()))) > duration;
 }
 
-@Pipe({name: 'relativeDate'})
+@Pipe({standalone: true, name: 'relativeDate'})
```
kosmoz: unnecessary default
```suggestion
@Pipe({name: 'relativeDate'})
```

frontend/cloud-ui/src/util/model.ts
```diff
@@ -8,7 +8,7 @@ export function isStale(model: BaseModel, duration: Duration = dayjs.duration({m
   return isOlderThan(model.createdAt, duration);
 }
 
-@Pipe({name: 'isStale'})
+@Pipe({standalone: true, name: 'isStale'})
```
kosmoz: unnecessary default
```suggestion
@Pipe({name: 'isStale'})
```

frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.ts
```diff
@@ -27,24 +31,33 @@ import {drawerFlyInOut} from '../animations/drawer';
     StatusDotComponent,
   ],
   templateUrl: './deployment-targets.component.html',
+  standalone: true,
```
kosmoz, line 43: unnecessary default
```suggestion
```

frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.html
```diff
@@ -144,15 +145,20 @@
...
               <td class="px-4 py-3 font-medium text-gray-900 whitespace-nowrap dark:text-white">
                 <div class="flex justify-end space-x-4">
                   @if (!dt.currentStatus) {
                     <button
                       type="button"
-                      (click)="showInstructions(instructionsModal)"
+                      (click)="showModal(instructionsModal)"
                       class="py-2 px-3 flex items-center text-sm font-medium text-center text-gray-900 focus:outline-none bg-white rounded-lg border border-gray-200 hover:bg-gray-100 hover:text-primary-700 focus:z-10 focus:ring-4 focus:ring-gray-200 dark:focus:ring-gray-700 dark:bg-gray-800 dark:text-gray-400 dark:border-gray-600 dark:hover:text-white dark:hover:bg-gray-700">
-                      Show Instructions
+                      Connect Deployment Target
```
kosmoz: Kind of pedantic, but since we're already in the process of updating wordings… It's easy to infer that this is about connecting a "deployment target". Wouldn't it make more sense to have a label like "Connect to …"? For example, "Connect to cluster" (for k8s) or "Connect to VM" or "VPS" or "server" or "machine" or "instance"?

frontend/cloud-ui/src/app/components/status-dot.ts
```diff
@@ -12,6 +12,7 @@ import {IsStalePipe} from '../../util/model';
       [class.bg-gray-500]="!deploymentTarget.currentStatus"></div>
   `,
   imports: [IsStalePipe],
+  standalone: true
```
kosmoz: unnecessary default
```suggestion
```

internal/handlers/deployments.go
pmig: Created a follow-up ticket for it to be able to have a separate page for deployments.

frontend/cloud-ui/src/app/deployment-targets/deployment-targets.component.ts
pmig: as discussed we are keeping the entities smaller, but can create a view model that fetches all the data.

PR #59, opened 2024-12-02
feat: add user authentication

kosmoz:


internal/auth/auth.go
```diff
@@ -0,0 +1,55 @@
...
+	defaultTokenExpiration = 24 * time.Hour
+)
+
+const (
+	UserNameKey  = "name"
+	UserEmailKey = "email"
+	OrgIdKey     = "org"
+)
+
+var JWTAuth = jwtauth.New("HS256", []byte("secret"), nil) // replace with secret key
```
christophenne: Is the key supposed to come from the environment? Could be added there in the `internal/env` package.

internal/db/user_accounts.go
```diff
@@ -0,0 +1,120 @@
...
+	} else {
+		return &userAccount, nil
+	}
+}
+
+// GetCurrentUser retrieves the user account ID from the context auth token (subject claim) and returns the
+// corresponding UserAccount
+//
+// TODO: this function should probably be moved to another module and maybe support some kind of result caching.
+func GetCurrentUser(ctx context.Context) (*types.UserAccount, error) {
```
christophenne, line 112: seems to be unused

internal/db/organization.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil
+	}
+}
+
+func GetOrganizationsForUser(ctx context.Context, userId string) ([]*types.Organization, error) {
+	db := internalctx.GetDb(ctx)
+	rows, err := db.Query(ctx, `
+		SELECT o.id, o.created_at, o.name
+			FROM UserAccount u
+			LEFT JOIN Organization_UserAccount j ON u.id = j.user_account_id
```
christophenne: I think these should be `INNER JOIN`s, otherwise when the user has no organizations, there would still be a row with `null`s, and then parsing the row to the struct would probably fail.

internal/handlers/auth.go
```diff
@@ -0,0 +1,83 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprint(w, "invalid username or password")
+	} else if orgs, err := db.GetOrganizationsForUser(r.Context(), user.ID); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		log.Warn("user login failed", zap.Error(err))
+	} else {
+		if len(orgs) < 1 {
+			w.WriteHeader(http.StatusInternalServerError)
+			log.Error("user has no organizations")
+		}
```
christophenne, line 46: maybe a warning if `len(orgs) > 1` makes sense as long as we don't support multi accounts

internal/types/data.go
```diff
@@ -46,3 +46,17 @@ type DeploymentTargetStatus struct {
 	Message            string    `db:"message" json:"message"`
 	DeploymentTargetId string    `db:"deployment_target_id" json:"-"`
 }
+
+type UserAccount struct {
+	Base
+	Email        string `db:"email" json:"email"`
+	PasswordHash []byte `db:"password_hash" json:"-"`
+	PasswordSalt []byte `db:"password_salt" json:"-"`
+	Password     string `db:"-" json:",omitempty"`
```
christophenne: I think this could also be `json:"-"`, right?

internal/db/organization.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, err
+	} else {
+		return result, nil
+	}
+}
+
+// GetCurrentOrg retrieves the organization_id from the context auth token and returns the corresponding Organization
+//
+// TODO: this function should probably be moved to another module and maybe support some kind of result caching.
+func GetCurrentOrg(ctx context.Context) (*types.Organization, error) {
```
christophenne, line 74: don't know if left in on purpose, but this is also unused

api/auth.go
```diff
@@ -0,0 +1,28 @@
+package api
```
christophenne, line 1: what's the idea behind the `api` package? should we move other types here too?

frontend/cloud-ui/src/app/services/auth.service.ts
```diff
@@ -0,0 +1,77 @@
...
+      throw new Error('token is null');
+    }
+  }
+
+  public logout(): Observable<void> {
+    this.token = null;
+    return of(undefined);
+  }
+
+  public httpInterceptor(): HttpInterceptorFn {
```
christophenne: probably removable in favor of the `tokenInterceptor`

frontend/cloud-ui/src/app/services/auth.service.ts
```diff
@@ -0,0 +1,77 @@
...
+      } else {
+        return next(req);
+      }
+    };
+  }
+}
+
+export const tokenInterceptor: HttpInterceptorFn = (req, next) => {
+  const auth = inject(AuthService);
+  if (req.url !== '/api/auth/login') {
```
christophenne, line 52: `api/auth/register` could also be excluded

internal/auth/auth.go
```diff
@@ -0,0 +1,55 @@
...
+func GenerateToken(user types.UserAccount, org types.Organization) (jwt.Token, string, error) {
+	now := time.Now()
+	claims := map[string]any{
+		jwt.IssuedAtKey:   now,
+		jwt.NotBeforeKey:  now,
+		jwt.ExpirationKey: now.Add(defaultTokenExpiration),
+		jwt.SubjectKey:    user.ID,
+		UserNameKey:       user.Name,
+		UserEmailKey:      user.Email,
+		OrgIdKey:          org.ID,
```
christophenne, line 40: Definitely fine for now, but just to keep long-term: When the organization(s) and possibly roles are in the token, the token should be short-lived and we should support some token refresh mechanism (i.e. token != session). Otherwise if access of a user to an organization is revoked, it will take a long time till that change is reflected.

frontend/cloud-ui/src/app/login/login.component.html
```diff
@@ -0,0 +1,69 @@
...
+            -->
+            <!-- TODO: enable "forgot password" feature if implemented
+            <a href="#" class="text-sm font-medium text-primary-600 hover:underline dark:text-primary-500"
+              >Forgot password?</a
+            >
+            -->
+          </div>
+          <button
+            type="submit"
+            class="w-full text-white bg-primary-600 hover:bg-primary-700 focus:ring-4 focus:outline-none focus:ring-primary-300 font-medium rounded-lg text-sm px-5 py-2.5 text-center dark:bg-primary-600 dark:hover:bg-primary-700 dark:focus:ring-primary-800">
```
christophenne, line 58: I think the primary flowbite classes are not working yet, that's why the sign in button looks like this:

Screenshot from 2024-12-04 14-46-05

internal/auth/auth.go
kosmoz: Yes definitely, this is only a shortcut. 

My long-term vision would be to keep the number of orgs in the token to one and have a session token that can be used to generate short-lived tokens.

frontend/cloud-ui/src/app/services/auth.service.ts
kosmoz: Will do in the follow up PR

api/auth.go
kosmoz: For authentication it made sense to have dedicated "DTO" struct definitions, that's what this package should be for. I wouldn't move the types here but at some point we might want to introduce a stricter separation of DB models and DTOs.

internal/db/organization.go
kosmoz: I created this (and the other one) because I though we will need them at some point. But I can also remove them for now, don't really mind

internal/types/data.go
kosmoz: In fact, I think the UserAccount is never returned by an api endpoint so all JSON annotations can be removed.

internal/db/user_accounts.go
kosmoz: see other comment

internal/auth/auth.go
kosmoz: good point, done

internal/db/organization.go
kosmoz: good catch

internal/types/data.go
kosmoz: I removed all of the "json" tags

frontend/cloud-ui/src/app/login/login.component.html
kosmoz: I added the primary color spec from flowbite. we can change it later if necessary.

PR #61, opened 2024-12-02
feat: small ui improvements, add demo data sql

pmig:


frontend/cloud-ui/src/app/applications/applications.component.ts
```diff
@@ -111,6 +111,7 @@ export class ApplicationsComponent {
         result = this.applications.update({
           id: val.id!,
           name: val.name!,
+          type: 'docker',
```
christophenne: why is this necessary, is it supposed to fix something? just asking because it looks like a source of error for the future

pmig: good point, I can do better there!

PR #78, opened 2024-12-05
build: improve makefile

kosmoz:


Makefile
```diff
@@ -22,20 +26,24 @@ lint: lint-go lint-frontend
 .PHONY: lint-fix
 lint-fix: lint-go-fix lint-frontend-fix
 
+node_modules: package-lock.json
+	npm install --no-save
```
pmig, line 30:
```suggestion
	npm ci
```

kosmoz: but it removes the entire node_modules every time…

pmig: good point

PR #83, opened 2024-12-05
feat: cloud agent

christophenne:


cmd/agent/main.go
```diff
@@ -0,0 +1,102 @@
...
+	"syscall"
+	"time"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+func main() {
+	accessKeyId := getFromEnvOrDie("GK_ACCESS_KEY_ID")
+	accessKeySecret := getFromEnvOrDie("GK_ACCESS_KEY_SECRET")
```
kosmoz, line 20: I would change the naming here to not resemble the naming used by AWS. This might lead to confusion. How about GLASSKUBE_TARGET_ID and GLASSKUBE_TARGET_SECRET?

cmd/agent/main.go
```diff
@@ -0,0 +1,102 @@
...
+				}
+				_ = resp.Body.Close()
+			}
+		}
+		time.Sleep(5 * time.Second)
+	}
+}
+
+func getFromEnvOrDie(arg string) string {
+	value := os.Getenv(arg)
```
kosmoz: better use `os.LookupEnv`

Dockerfile.agent
```diff
@@ -0,0 +1,25 @@
...
+
+COPY cmd/ cmd/
+# doesn't exist (yet?)
+# COPY pkg/ pkg/
+COPY internal/ internal/
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o agent ./cmd/agent/
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+# FROM cgr.dev/chainguard/static:latest@sha256:5ff428f8a48241b93a4174dbbc135a4ffb2381a9e10bdbbc5b9db145645886d5
```
kosmoz:
```suggestion
```

Dockerfile.agent
```diff
@@ -0,0 +1,25 @@
...
+FROM alpine:3.20
```
kosmoz: maybe consider https://hub.docker.com/_/docker as a base image

cmd/agent/main.go
```diff
@@ -0,0 +1,102 @@
...
+		sigint := make(chan os.Signal, 1)
+		signal.Notify(sigint, syscall.SIGTERM, syscall.SIGINT)
+		<-sigint
+		fmt.Println("ok bye")
+		os.Exit(0)
+	}()
+
+	client := http.Client{}
+
+	for {
```
kosmoz: Something like this could be implemented to ensure the agent is not interrupted during an operation:
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	go func() {
		sigint := make(chan os.Signal, 1)
		signal.Notify(sigint, syscall.SIGTERM, syscall.SIGINT)
		<-sigint
		cancel()
	}()

	client := http.Client{}

	for ctx.Err() == nil {
```

cmd/agent/main.go
```diff
@@ -0,0 +1,102 @@
...
+		EncoderConfig:     encoderCfg,
+		OutputPaths: []string{
+			"stderr",
+		},
+		ErrorOutputPaths: []string{
+			"stderr",
+		},
+	}
+
+	return zap.Must(config.Build())
```
kosmoz: 2 things:
1. Is there a reason why you don't use NewDevelopmentConfig()/NewProductionConfig()?
2. Wouldn't it be better to use `Development: true`, at least for now?

I think something like this would be quite similar to the current logger but far less config:
```go
	config := zap.NewDevelopmentConfig()
	config.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
	return zap.Must(config.Build())
```

internal/handlers/agent.go
```diff
@@ -0,0 +1,143 @@
...
+	accessKeyId := r.URL.Query().Get("accessKeyId")
+	accessKeySecret := r.URL.Query().Get("accessKeySecret")
+
+	if accessKeyId == "" || accessKeySecret == "" {
+		log.Warn("connect endpoint called without accessKeyId or accessKeySecret", zap.String("accessKeyId", accessKeyId))
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	deploymentTarget, err := db.GetDeploymentTargetUnauthenticated(ctx, accessKeyId)
```
kosmoz: I think it would be better to remove the "authentication" from the DB layer and move it to the API layer

internal/handlers/agent.go
```diff
@@ -0,0 +1,143 @@
...
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	} else if deploymentTarget.AccessKeySalt == nil || deploymentTarget.AccessKeyHash == nil {
+		log.Warn("deployment target does not have access key salt and hash configured",
+			zap.String("deploymentTargetId", deploymentTarget.ID))
+		w.WriteHeader(http.StatusUnauthorized)
+	} else if err := security.VerifyAccessKey(
+		*deploymentTarget.AccessKeySalt, *deploymentTarget.AccessKeyHash, accessKeySecret); err != nil {
+		log.Error("failed to verify access key secret", zap.Error(err))
+		w.WriteHeader(http.StatusUnauthorized)
```
kosmoz: maybe move to a middleware

internal/handlers/agent.go
```diff
@@ -0,0 +1,143 @@
...
+	} else {
+		u = u.JoinPath("/api/resources")
+		return u.String(), nil
+	}
+}
+
+func downloadResources(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := internalctx.GetLogger(ctx)
+	authHeader := r.Header.Get("Authorization")
```
kosmoz: there is a `r.BasicAuth()` function for this

kosmoz: also, maybe move to a middleware

cmd/agent/main.go
```diff
@@ -0,0 +1,95 @@
...
+
+			if resp, err := client.Do(req); err != nil {
+				logger.Error("failed to execute request", zap.Error(err))
+			} else {
+				if body, err := io.ReadAll(resp.Body); err != nil {
+					logger.Error("failed to read response body", zap.Error(err))
+				} else {
+					fmt.Fprintf(os.Stderr, "---response body: \n%v\n---\n", string(body))
+
+					err = os.WriteFile("/tmp/compose.yaml", body, 0644)
```
kosmoz, line 49: better use os.TempDir()

cmd/agent/main.go
```diff
@@ -0,0 +1,95 @@
...
+
+	client := http.Client{}
+
+	for ctx.Err() == nil {
+		if req, err := http.NewRequest("GET", endpoint, nil); err != nil {
+			logger.Error("failed to create request", zap.Error(err))
+		} else {
+			auth := fmt.Sprintf("%s:%s", accessKeyId, accessKeySecret)
+			encodedAuth := base64.StdEncoding.EncodeToString([]byte(auth))
+			req.Header.Set("Authorization", "Basic "+encodedAuth)
```
kosmoz: better use req.SetBasicAuth()

cmd/agent/main.go
```diff
@@ -0,0 +1,95 @@
...
+				if body, err := io.ReadAll(resp.Body); err != nil {
+					logger.Error("failed to read response body", zap.Error(err))
+				} else {
+					fmt.Fprintf(os.Stderr, "---response body: \n%v\n---\n", string(body))
+
+					err = os.WriteFile("/tmp/compose.yaml", body, 0644)
+					if err != nil {
+						logger.Error("failed to write temp file", zap.Error(err))
+					} else if out, err := exec.Command(
+						"docker", "compose", "-f", "/tmp/compose.yaml", "up", "-d").CombinedOutput(); err != nil {
```
kosmoz: maybe add --quiet-pull for less noisy output

internal/db/deployments.go
```diff
@@ -74,6 +75,29 @@ func GetLatestDeploymentForDeploymentTarget(ctx context.Context, deploymentTarge
 	}
 }
 
+func GetLatestDeploymentComposeFile(ctx context.Context, deploymentTargetId string, orgId string) ([]byte, error) {
+	db := internalctx.GetDb(ctx)
+	if rows, err := db.Query(ctx, `
+		SELECT av.compose_file_data
+		FROM Deployment d
+		INNER JOIN ApplicationVersion av ON d.application_version_id = av.id
+		INNER JOIN DeploymentTarget dt ON d.deployment_target_id = d.id
```
kosmoz: found a bug :slightly_smiling_face: 
```suggestion
		INNER JOIN DeploymentTarget dt ON d.deployment_target_id = dt.id
```

cmd/agent/main.go
```diff
@@ -0,0 +1,95 @@
...
+			} else {
+				if body, err := io.ReadAll(resp.Body); err != nil {
+					logger.Error("failed to read response body", zap.Error(err))
+				} else {
+					fmt.Fprintf(os.Stderr, "---response body: \n%v\n---\n", string(body))
+
+					err = os.WriteFile("/tmp/compose.yaml", body, 0644)
+					if err != nil {
+						logger.Error("failed to write temp file", zap.Error(err))
+					} else if out, err := exec.Command(
```
kosmoz, line 52: maybe not so important now, but wouldn't it be even simpler to just copy the request body straight to the command stdin?

christophenne: Good idea, I will do it in the follow up PR (also the renaming above)

PR #84, opened 2024-12-05
feat(backend): add mail sending capabilities

kosmoz:
This is a basic first version of the email feature, but in order to keep PRs smaller, I will add ~HTML templates and other~ more stuff in another PR once this one gets merged.

The `MAILER_TYPE` env variable can be changed to `"ses"`, to change the mailer imlementation to the AWS SES client. This is convenient when deployed on AWS because no SMTP credentials or other config are required. But it is not convenient for dev, so there is an alternative SMTP implementation.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -3,17 +3,12 @@ package main
 import (
 	"context"
 	"fmt"
-	"io/fs"
 	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
 
 	"github.com/glasskube/cloud/internal/server"
-	"github.com/go-chi/chi/v5/middleware"
-
-	"github.com/glasskube/cloud/internal/frontend"
-	"github.com/go-chi/chi/v5"
 )
 
 func main() {
@@ -25,18 +20,6 @@ func main() {
 	}
 	defer func() { _ = s.Shutdown() }()
 
-	router := chi.NewRouter()
-	router.Use(
-		// Handles panics
-		middleware.Recoverer,
-		// Reject bodies larger than 1MiB
-		middleware.RequestSize(1048576),
-	)
-	router.Mount("/api", server.ApiRouter(s))
-	router.With(
-		middleware.Compress(5, "text/html", "text/css", "text/javascript"),
-	).Handle("/*", StaticFileHandler())
-
 	go func() {
 		sigint := make(chan os.Signal, 1)
 		signal.Notify(sigint, syscall.SIGTERM, syscall.SIGINT)
@@ -47,20 +30,7 @@ func main() {
 
 	addr := ":8080"
 	s.GetLogger().Sugar().Infof("listen on %v", addr)
-	if err := http.ListenAndServe(addr, router); err != nil {
+	if err := http.ListenAndServe(addr, server.NewRouter(s)); err != nil {
 		panic(err)
-	}
-}
-
-func StaticFileHandler() http.HandlerFunc {
-	fsys := frontend.BrowserFS()
-	server := http.FileServer(http.FS(fsys))
-	return func(w http.ResponseWriter, r *http.Request) {
-		// check if the requested file exists and use index.html if it does not.
-		if _, err := fs.Stat(fsys, r.URL.Path[1:]); err != nil {
-			http.StripPrefix(r.URL.Path, server).ServeHTTP(w, r)
-		} else {
-			server.ServeHTTP(w, r)
-		}
 	}
 }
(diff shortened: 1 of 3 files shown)


internal/server/main.go
```diff
@@ -45,36 +65,47 @@ func Init() error {
...
-	return nil
 }
 
-func Shutdown() error {
-	if instance == nil {
-		return errors.New("server not yet initialized")
-	}
-	instance.dbPool.Close()
-	return nil
+func (s *server) GetDbPool() *pgxpool.Pool {
```
christophenne, line 80: Seems to me that `GetDbPool` and `GetMailer` could stay private, they are only used in the `server` package.

internal/mail/mail.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+func Subject(subject string) mailOpt {
+	return func(mail *Mail) {
+		mail.Subject = subject
+	}
+}
+
+func HtmlBody(body string) mailOpt {
```
christophenne, line 29: unused but probably left in on purpose

internal/mail/context.go
```diff
@@ -0,0 +1,19 @@
+package mail
+
+import "context"
+
+type contextKey int
+
+const key contextKey = iota
+
+func AddToContext(ctx context.Context, mailer Mailer) context.Context {
```
christophenne: Is there a reason why mailer context handling is not part of the `internal/context` package where we also do the most other context handling?

internal/emailtemplates/templates.go
```diff
@@ -0,0 +1,42 @@
+package emailtemplates
```
christophenne: extremely unimportant, but I would call it `mailtemplates` :sweat_smile:

cmd/main.go
```diff
@@ -16,18 +17,22 @@ import (
...
-		panic(err)
+	ctx := context.Background()
+
+	s, err := server.New(ctx)
+	if err != nil {
+		panic(fmt.Errorf("failed to init server: %w", err))
 	}
+	defer func() { _ = s.Shutdown() }()
+
 	router := chi.NewRouter()
```
christophenne: not related to this PR, but it just occurred to me that even this router should also move to the `server` package, such that the `cmd` only starts and stops it.

kosmoz: I thought about this as well and IMO it would be even better to completely separate the server and router. In fact, I considered the following module structure: 

`application`/`service`/`svc` (not sure about the name) -- responsible for initializing Services, such as DBPool, Mailer, Logger
`server` -- responsible to start and stop the HTTP server
`router`/`routing` -- just the routes

BTW, the reason why the getter methods on the server are public (other comment) is because I kind of anticipated a change like this, but in the end I didn't want to go through with it just yet :sweat_smile: 

If you agree with these changes, I can do the refactoring, but would probably do it in a separate PR. 

Also, since the env parsing is already getting kind of complicated, I though it might make sense to migrate to https://github.com/spf13/viper already, what do you think?

internal/mail/context.go
kosmoz: Good question… I think objectively it doesn't really matter and it's rather a question of preference. IMO `mailer.FromContext(ctx)` reads easier than `internalctx.GetMailer(ctx)`, so that's what I went with. 

I understand that it's questionable to have both styles in the same codebase, so I can move it to internalctx, but maybe we can consider refactoring to the (subjectively, IMO) more readable style in a separate PR?

internal/mail/mail.go
kosmoz: Yes, I wanted to offer a bit more than necessary so we don't have to touch the mailer package that often (hopefully)

internal/server/main.go
kosmoz: see other comment

internal/emailtemplates/templates.go
kosmoz: fine by me :smile:

internal/mail/context.go
christophenne: > move it to internalctx, but maybe we can consider refactoring to the (subjectively, IMO) more readable style in a separate PR

sounds good to me

cmd/main.go
christophenne: No strong feelings, it can be refactored to that structure, but I don't think it has a high priority at all.

internal/mail/context.go
kosmoz: ok moved to internalctx for now

cmd/main.go
kosmoz: OK, I moved all routing to the router.go file. I think that makes sense for now.

PR #94, opened 2024-12-09
feat: add user role

kosmoz:


sql/init_db.sql
```diff
@@ -1,89 +1,91 @@
...
   message TEXT NOT NULL
 );
 
 CREATE INDEX IF NOT EXISTS fk_DeploymentTargetStatus_deployment_target_id ON DeploymentTargetStatus (deployment_target_id);
 
 CREATE TABLE IF NOT EXISTS Deployment (
-  id UUID PRIMARY KEY DEFAULT gen_random_uuid (),
+  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
   created_at TIMESTAMP DEFAULT current_timestamp,
   deployment_target_id UUID NOT NULL REFERENCES DeploymentTarget (id),
```
christophenne, line 77: if the idea was to put `ON DELETE CASCADE` everywhere, here are two missing

internal/db/deployment_targets.go
```diff
@@ -100,7 +134,7 @@ func CreateDeploymentTarget(ctx context.Context, dt *types.DeploymentTarget) err
 	if err != nil {
 		return fmt.Errorf("could not save DeploymentTarget: %w", err)
 	} else {
-		*dt = result
+		dt.DeploymentTarget = result
 		return nil
 	}
 }
```
christophenne, line 148: github won't let me make a suggestion for unmodified parts. in line 156 the expression should be changed to `deploymentTargetOutputExprBase` for the deployment target update to work

internal/db/deployment_targets.go
```diff
@@ -77,21 +93,39 @@ func GetDeploymentTarget(ctx context.Context, id string, orgId *string) (*types.
...
+			return err
+		} else {
+			dt.OrganizationID = orgId
+		}
+	}
+	if dt.CreatedBy == nil {
+		if userId, err := auth.CurrentUserId(ctx); err != nil {
+			return err
+		} else {
+			dt.CreatedBy = &types.UserAccount{ID: userId}
```
christophenne, line 112: we could also fill the rest of the fields from what is available in the token, otherwise when creating a deployment target via the UI, the created by column is empty (which is just a minor bug)

christophenne: or we simply don't show the created by in the UI for now

sql/init_db.sql
kosmoz: The idea was to be able to delete an organization with associated apps and deployment targets, as long as there are no deployments (for safety). If there are deployments I think it makes sense to require deleting those first to prevent accidental deletion

internal/db/deployment_targets.go
kosmoz: Fixed using a `WITH` query

internal/db/deployment_targets.go
kosmoz: update query, now the user is included after upate

PR #100, opened 2024-12-10
feat(frontend): global error handling

christophenne:


angular.json
```diff
@@ -32,7 +32,8 @@
               }
             ],
             "styles": [
-              "frontend/cloud-ui/src/styles.scss"
+              "frontend/cloud-ui/src/styles.scss",
+              "node_modules/ngx-toastr/toastr.css"
```
kosmoz: is this even needed if we use the custom toast component?

christophenne: Yes, for the positioning class. I tried to use the tailwind classes instead but the toastr seems to support only one class and not multiple.

kosmoz: ok just tried and it is needed, but I think it would be better to import it in styles.scss instead because that's what we do for other css dependencies already

frontend/cloud-ui/src/app/services/error-toast.interceptor.ts
```diff
@@ -0,0 +1,37 @@
...
+            toastr.error(err.statusText, 'An error occurred', GLOBAL_ERROR_TOAST_CONFIG);
+          }
+        }
+      },
+    })
+  );
+};
+
+function ignoreGlobalError(req: HttpRequest<any>, err: HttpErrorResponse): boolean {
+  if (err.status >= 400 && err.status < 500) {
```
kosmoz, line 33: why only for 500?

christophenne: The idea was that client side errors like 400 are more use case specific. For example in the deployment targets view, the latest-deployment call may return 404 which is "ok" for this use case. Same for example for the login, where I might want to display "wrong username or password" directly in the form and not in the toast somewhere else on the page.

kosmoz: I see. network errors (status code "0") are also shown, so it's OK for me

frontend/cloud-ui/src/app/services/error-toast.interceptor.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export const errorToastInterceptor: HttpInterceptorFn = (req, next) => {
+  const toastr = inject(ToastrService);
+  return next(req).pipe(
+    tap({
+      error: (err) => {
+        if (err instanceof HttpErrorResponse) {
+          if (!ignoreGlobalError(req, err)) {
+            toastr.error(err.statusText, 'An error occurred', GLOBAL_ERROR_TOAST_CONFIG);
```
kosmoz: the toast has some weird visual artifacts. those are because the library adds a toast-error class to the element which is part of the toastr library css. this can be prevented in two ways:
1. only copy the necessary parts of the css directly into our css
2. use `toastr.show(..., '')` (empty type) instead of `toastr.error(...)`

christophenne: not sure if its fixed, I didn't see the visual artifacts.

kosmoz: weird, but now it works

PR #101, opened 2024-12-10
feat(frontend): integrate sentry

christophenne:
* [ ] token
* [ ] remove test stuff from main

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -23,9 +23,3 @@ jobs:
         run: golangci-lint run --out-format=colored-line-number
       - name: Build application
         run: make build
-      - name: Sentry Sourcemaps (just to test)
-        run: make sentry-upload-sourcemaps
-        env:
-          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
-      - name: Delete Sourcemaps (just to test)
-        run: make delete-sourcemaps
--- a/frontend/cloud-ui/src/app/app.config.ts
+++ b/frontend/cloud-ui/src/app/app.config.ts
@@ -18,13 +18,6 @@ export const appConfig: ApplicationConfig = {
       provide: ErrorHandler,
       useValue: Sentry.createErrorHandler(),
     },
-    {
-      provide: Sentry.TraceService,
-      deps: [Router],
-    },
-    provideAppInitializer(() => {
-      inject(Sentry.TraceService);
-    }),
     provideZoneChangeDetection({eventCoalescing: true}),
     provideRouter(routes),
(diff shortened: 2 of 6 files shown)


.github/workflows/main.yaml
```diff
@@ -23,3 +23,9 @@ jobs:
         run: golangci-lint run --out-format=colored-line-number
       - name: Build application
         run: make build
+      - name: Sentry Sourcemaps (just to test)
```
christophenne: should be removed before merging, I just wanted to test it but it doesn't work yet because the token is still missing and I cannot add it

frontend/cloud-ui/src/app/app.config.ts
```diff
@@ -1,12 +1,30 @@
...
   providers: [
+    {
+      provide: ErrorHandler,
+      useValue: Sentry.createErrorHandler(),
+    },
+    {
+      provide: Sentry.TraceService,
+      deps: [Router],
+    },
+    provideAppInitializer(() => {
```
christophenne: official docs still use the outdated APP_INITIALIZER, but according to https://github.com/getsentry/sentry-javascript/pull/14398 this is the new way

.github/workflows/main.yaml
kosmoz: I think we can worry about source maps later, because there are still some open issues around them:
1. it should really only be done when building the docker image, because that's what will be deployed in the end
2. Source maps work best if you can relate an event to a version, which we currently can't do
3. I think it might be easier to use an angular builder that allows specifying custom esbuild plugins and use the sentry esbuild plugin (which can also delete source maps after uploading) https://www.npmjs.com/package/@angular-builders/custom-esbuild (`latest` does not support v19 yet) https://docs.sentry.io/platforms/javascript/guides/angular/sourcemaps/uploading/esbuild/

frontend/cloud-ui/src/app/app.config.ts
kosmoz: I think the docs are up to date now: https://docs.sentry.io/platforms/javascript/guides/angular/ :slightly_smiling_face: 

But from what I understand, to make the TraceService work, you need to add the integration in the sentry config as well… that is if you want to use it at all, which I think we might also skip for now and only use error capturing

.github/workflows/main.yaml
kosmoz: What I mean by this is I think they should be removed from at least the actions workflow files for now

frontend/cloud-ui/src/app/app.config.ts
christophenne: ok not docs, but the guide in the sentry UI

PR #145, opened 2024-12-16
feat: email verification

christophenne:
:warning:  needs DB migration

```
ALTER TABLE UserAccount ADD COLUMN email_verified_at TIMESTAMP;
```

To update all existing users to be verified:

```
UPDATE UserAccount SET email_verified_at = current_timestamp WHERE email_verified_at IS NULL;
```

### DIFF
--- a/internal/mailtemplates/templates/verify-email-registration.html
+++ b/internal/mailtemplates/templates/verify-email-registration.html
@@ -13,7 +13,8 @@
         {{else}}
         <p>Hi,</p>
         {{end}} Welcome to <a href="https://glasskube.cloud/">glasskube.cloud</a>! Please
-        <a href="{{.Host}}/verify?jwt={{.Token | QueryEscape}}">click this link</a> to verify that this is your email address.
+        <a href="{{.Host}}/verify?jwt={{.Token | QueryEscape}}">click this link</a> to verify that this is your email
+        address.
 
         <p>{{template "fragments/signature.html"}}</p>
       </main>


internal/handlers/auth.go
```diff
@@ -91,11 +93,21 @@ func authRegisterHandler(w http.ResponseWriter, r *http.Request) {
...
 
+		_, token, err := auth.GenerateVerificationTokenValidFor(
+			userAccount,
+			types.OrganizationWithUserRole{Organization: *org, UserRole: types.UserRoleVendor},
+			env.InviteTokenValidDuration(),
+		)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
```
kosmoz: I think this should either be moved to inside RunTx or not cause a 500 error, because in case of a 500 error I would expect the changes to the DB to be rolled back

internal/handlers/settings.go
```diff
@@ -34,6 +37,9 @@ func updateUserSettingsHandler(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 		}
 	}
+	if body.EmailVerified && user.EmailVerifiedAt == nil {
+		user.EmailVerifiedAt = util.PtrTo(time.Now())
+	}
```
kosmoz: maybe only allow this if email_verified in the JWT is already true?

PR #155, opened 2024-12-16
feat(backend): add db migrations

kosmoz:
For the most part the migrations would work, because we use `CREATE … IF EXISTS …` everywhere, but such a thing does not exist for `CREATE TYPE …` statements. 

Although there are some more or less sketchy workarounds for this, I think the most responsible solution is to just create a new database and do migrations for everything from now on…


cmd/cloud/migrate/up/main.go
```diff
@@ -0,0 +1,16 @@
...
+	"context"
+
+	"github.com/glasskube/cloud/internal/migrations"
+	"github.com/glasskube/cloud/internal/svc"
+	"github.com/glasskube/cloud/internal/util"
+)
+
+func main() {
+	ctx := context.Background()
+	registry := util.Require(svc.NewDefault(ctx))
```
christophenne, line 13: We could make a more minimal registry, because for the two migration commands only the logger is required, but mailer and the unused pgx db pool are still instantiated, and the configs need to be set too (mailer, host for example). That on the other hand needs some different concept to make certain environment variables only mandatory if certain "services" are needed. And that would need a different concept for environment / config handling like already discussed once. So it can definitely stay like that for now.

internal/migrations/migrate.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/glasskube/cloud/internal/env"
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database/postgres"
+	"github.com/golang-migrate/migrate/v4/source/iofs"
+	_ "github.com/jackc/pgx/v5/stdlib"
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+)
+
+//go:embed *
```
christophenne: I think the sqls could be put into a subdirectory, but no strong feelings.

cmd/cloud/migrate/up/main.go
christophenne: (if the default pgx pool is even unused – if it is actually used that would be quite magic too)

kosmoz: I tried to use the pgxpool initially, but it is not compatible, unfortunately. go-migrate only works with standard sql.DB instances, but at least pgx supports those too.

The svc registry is only needed for the logger actually, but since these scripts are only intended for dev use, I honestly wouldn't bother adding any more code here than absolutely necessary.

internal/migrations/migrate.go
kosmoz: me too, moved to ./sql

PR #169, opened 2024-12-17
feat(backend): add sentry

kosmoz:


internal/routing/routing.go
```diff
@@ -47,6 +48,7 @@ func ApiRouter(logger *zap.Logger, db *pgxpool.Pool, mailer mail.Mailer) http.Ha
 		r.Group(func(r chi.Router) {
 			r.Use(jwtauth.Verifier(auth.JWTAuth))
 			r.Use(jwtauth.Authenticator(auth.JWTAuth))
+			r.Use(middleware.SentryUser)
```
christophenne, line 51: I think this could also be used in the JWT-authenticated agent routes, only difference is that the subject is the deployment target id. Not sure if that brings significant benefit, but maybe.

kosmoz: good idea, done

PR #179, opened 2024-12-18
feat: add entity sorting

pmig:


internal/db/deployment_targets.go
```diff
@@ -65,7 +65,8 @@ func GetDeploymentTargets(ctx context.Context) ([]types.DeploymentTargetWithCrea
 	if rows, err := db.Query(ctx,
 		"SELECT"+deploymentTargetWithStatusOutputExpr+deploymentTargetFromExpr+
 			"WHERE dt.organization_id = @orgId "+
-			"AND (dt.created_by_user_account_id = @userId OR @userRole = 'vendor')",
+			"AND (dt.created_by_user_account_id = @userId OR @userRole = 'vendor') "+
+			"ORDER BY u.name", // does not work?
```
kosmoz: if it doesn't work, please remove it or fix it

if it does work, please remove the comment, as it might be confusing :slightly_smiling_face:

PR #200, opened 2024-12-20
feat: release info and sentry sourcemaps

kosmoz:


frontend/cloud-ui/src/main.ts
```diff
@@ -25,6 +27,21 @@ if (environment.posthogToken) {
...
+      (cr) => {
+        if (cr !== null) {
+          if (cr.$set === undefined) {
+            cr.$set = {};
+          }
+          if (cr.$set_once === undefined) {
+            cr.$set_once = {};
+          }
+          cr.$set['version'] = buildConfig.version;
+          cr.$set_once['version'] = buildConfig.version;
```
christophenne: why does it need to be both `set` and `set_once`? doesn't `set_once` only set it if it has no value yet?

kosmoz: you're right. I changed it to intial_version, same as in the packagemanager project

PR #215, opened 2025-01-02
fix(ui): use correct application version in deploy form

christophenne:


frontend/cloud-ui/src/app/components/installation-wizard/installation-wizard.component.ts
```diff
@@ -63,9 +63,13 @@ export class InstallationWizardComponent implements OnInit, OnDestroy {
   private readonly destroyed$ = new Subject<void>();
 
   ngOnInit() {
-    this.deployForm.controls.applicationId.valueChanges
-      .pipe(takeUntil(this.destroyed$))
-      .subscribe((it) => this.updatedSelectedApplication(it!));
+    this.deployForm.controls.applicationId.valueChanges.pipe(takeUntil(this.destroyed$)).subscribe(async (it) => {
```
kosmoz: I realize that `updatedSelectedApplication` was already an async function before, but I feel that using async/await in a subscriber seems like a bad idea as it's kind of hard to reason about what happens if two events happen close to each other. 

Maybe a better way would be to use switchMap to get the applications and then no async is needed.

frontend/cloud-ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -211,13 +219,21 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
...
-  private readonly applicationIdChange$ = this.deployForm.controls.applicationId.valueChanges.subscribe((it) =>
-    this.updatedSelectedApplication(it!!)
-  );
-
-  ngOnDestroy(): void {
-    this.applicationIdChange$.unsubscribe();
-    this.destroyed$.complete();
+  private registerDeployFormChanges() {
+    this.deployForm.controls.applicationId.valueChanges.pipe(takeUntil(this.destroyed$)).subscribe(async (it) => {
+      await this.updatedSelectedApplication(it!!);
```
kosmoz:
```suggestion
      await this.updatedSelectedApplication(it!);
```

frontend/cloud-ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -211,13 +219,21 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
...
+      if (this.selectedApplication && (this.selectedApplication.versions || []).length > 0) {
```
kosmoz:
```suggestion
      if (this.selectedApplication && (this.selectedApplication.versions ?? []).length > 0) {
```

frontend/cloud-ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -211,13 +219,21 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
...
 
-  private readonly applicationIdChange$ = this.deployForm.controls.applicationId.valueChanges.subscribe((it) =>
-    this.updatedSelectedApplication(it!!)
-  );
-
-  ngOnDestroy(): void {
-    this.applicationIdChange$.unsubscribe();
-    this.destroyed$.complete();
+  private registerDeployFormChanges() {
+    this.deployForm.controls.applicationId.valueChanges.pipe(takeUntil(this.destroyed$)).subscribe(async (it) => {
```
kosmoz: see other comment about async/await

frontend/cloud-ui/src/app/components/installation-wizard/installation-wizard.component.ts
christophenne: Yes, now tried it with a more rxjs-y way

PR #220, opened 2025-01-02
ci: rewrite build workflow to accomodate sentry upload

kosmoz:


.github/workflows/build-server.yaml
```diff
@@ -0,0 +1,81 @@
...
+        run: npm ci
+
+      - name: Lint frontend
+        run: npm run lint
+      - name: Lint go project
+        run: golangci-lint run --out-format=colored-line-number
+
+      - name: Build frontend
+        if: ${{ !startsWith(github.ref, 'refs/tags/') }}
+        run: npm run build:prod -- --source-map=true
```
christophenne: Is `--source-map=true` even necessary here?

christophenne: But I guess it doesn't matter because it's never deployed anyway.

kosmoz: you're right on both accounts :smile: I removed it to prevent confusion

PR #227, opened 2025-01-03
feat: add helm deployment support for deployment targets and deployments

kosmoz:


frontend/cloud-ui/src/app/components/onboarding-wizard/onboarding-wizard.component.html
```diff
@@ -542,7 +542,7 @@ <h3 class="text-xl font-semibold text-gray-900 dark:text-white">Connect Customer
                         Execute the following in the terminal:
                         @if (createdDeploymentTarget?.id) {
                           <app-connect-instructions
-                            [deploymentTargetId]="createdDeploymentTarget!.id!"></app-connect-instructions>
```
christophenne, line 545: Additionally in the onboarding wizard, the selected deployment type must be used instead of the hardcoded docker, otherwise deployment creation fails with "application and deployment target must have the same type"

frontend/cloud-ui/src/app/deployments/deployment-targets.component.html
```diff
@@ -117,7 +117,7 @@
               <td
                 class="px-4 py-3 font-medium text-gray-900 whitespace-nowrap dark:text-white flex justify-end space-x-2">
                 @if (!(auth.hasRole('vendor') && dt.createdBy?.userRole === 'customer')) {
-                  @if (dt.currentStatus) {
+                  @if (dt.currentStatus || true) {
```
christophenne:
```suggestion
                  @if (dt.currentStatus) {
```

frontend/cloud-ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -57,30 +71,57 @@ import {filteredByFormControl} from '../../util/filter';
...
+  selectedApplication?: Application | null;
+
+  readonly filterForm = new FormGroup({
     search: new FormControl(''),
   });
 
+  readonly editForm = new FormGroup({
+    id: new FormControl<string | undefined>(undefined),
+    name: new FormControl('', Validators.required),
+    type: new FormControl({value: '', disabled: true}, Validators.required),
```
christophenne: This could be typed now with the new `DeploymentType`.

internal/handlers/applications.go
```diff
@@ -263,29 +265,36 @@ func updateApplicationVersion(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func getApplicationVersionComposeFile(w http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
-	log := internalctx.GetLogger(ctx)
-	application := internalctx.GetApplication(ctx)
-	applicationVersionId := r.PathValue("applicationVersionId")
-	// once performance becomes more important, do not load the whole application but only the requested version
```
christophenne, line 271: I think now this handler doesn't need to load the application into the context anymore, so it could be removed to save a query. On the other hand, it might be a bit difficult to express in the router config, so I don't know.

internal/types/data.go
```diff
@@ -59,8 +59,10 @@ func (av ApplicationVersion) Validate(deplType DeploymentType) error {
 
 type Deployment struct {
 	Base
-	DeploymentTargetId   string `db:"deployment_target_id" json:"deploymentTargetId"`
-	ApplicationVersionId string `db:"application_version_id" json:"applicationVersionId"`
+	DeploymentTargetId   string  `db:"deployment_target_id" json:"deploymentTargetId"`
+	ApplicationVersionId string  `db:"application_version_id" json:"applicationVersionId"`
+	ReleaseName          *string `db:"release_name" json:"releaseName"`
```
christophenne, line 64: I think this is never set (yet) (I think), is it on purpose?

internal/handlers/applications.go
kosmoz: Yeah, I've been thinking about that too but I came to the conclusion that it would be better to keep the router simple because it's not an expensive query. at least for now

christophenne: Fine to me, but please put the comment back in :)

kosmoz: Nevermind I updated the router config because I think it's not that bad after all

internal/types/data.go
kosmoz: thanks, I forgot about that :see_no_evil:

frontend/cloud-ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -150,7 +151,7 @@ export class DeploymentTargetsComponent implements OnInit, AfterViewInit, OnDest
 
   ngAfterViewInit() {
     if (this.fullVersion) {
-      combineLatest([this.filteredApplications$, this.deploymentTargets$])
+      combineLatest([this.applications$, this.deploymentTargets$])
```
christophenne, line 154: Changed back from `filteredApplications$` on purpose or by accident? I think it was correct.

christophenne: Nevermind, got it.

PR #232, opened 2025-01-07
feat: add helm deployment support for applications

christophenne:


internal/handlers/applications.go
```diff
@@ -155,11 +156,47 @@ func createApplicationVersion(w http.ResponseWriter, r *http.Request) {
...
+		} else {
+			applicationVersion.ComposeFileData = data
+		}
+	} else {
+		if data := readFile(w, r, "valuesfile", false); data != nil {
+			applicationVersion.ValuesFileData = data
+		}
+		if data := readFile(w, r, "templatefile", false); data != nil {
+			applicationVersion.TemplateFileData = data
+		}
```
kosmoz: If I understand correctly, this will continue to save the application version even in case there was an error. Even if the field is not required in the form, I think nothing should be saved if the client receives an error status code

frontend/cloud-ui/src/util/forms.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function toggleControls(formGroup: FormGroup, enabled: boolean) {
+  for (let controlsKey in formGroup.controls) {
+    if (enabled) {
+      formGroup.controls[controlsKey].enable();
+    } else {
+      formGroup.controls[controlsKey].disable();
+    }
+  }
```
kosmoz: wouldn't it be better to enable/disable the entire FormGroup? According to the documentation, this should also enable/disable all children.

frontend/cloud-ui/src/app/components/onboarding-wizard/onboarding-wizard.component.ts
```diff
@@ -75,10 +78,65 @@ export class OnboardingWizardComponent implements OnInit, OnDestroy {
         Validators.required
       ),
     }),
+    kubernetes: new FormGroup({
+      name: new FormControl<string>(
+        {
+          value: '',
+          disabled: true,
```
kosmoz: also here, I think it should be easier to disable the entire group, but this would need to be done in the component constructor/ngOnInit because the FormGroup constructor does not have this option.

frontend/cloud-ui/src/app/components/onboarding-wizard/onboarding-wizard.component.ts
```diff
@@ -141,30 +242,60 @@ export class OnboardingWizardComponent implements OnInit, OnDestroy {
...
+            };
+          } else {
+            name = this.applicationForm.controls.kubernetes.controls.name.value!;
+            version = {
+              name: this.applicationForm.controls.kubernetes.controls.versionName.value!,
+              chartType: this.applicationForm.controls.kubernetes.controls.chartType.value!,
+              chartName:
+                this.applicationForm.controls.kubernetes.controls.chartType.value === 'repository'
+                  ? this.applicationForm.controls.kubernetes.controls.chartName.value!
+                  : undefined,
```
kosmoz: I think the ternary here is not necessary, because reactive forms removes disabled controls from the form value. So, if chartType is 'repository', chartName *should* always be undefined anyways.

christophenne: Apparently its only undefined if you access the value via the form group, which I do now.

kosmoz: ah yes you are right

internal/types/data.go
```diff
@@ -12,11 +15,45 @@ type Application struct {
...
+	Name      string    `db:"name" json:"name"`
+
+	ChartType    *HelmChartType `db:"chart_type" json:"chartType,omitempty"`
+	ChartName    *string        `db:"chart_name" json:"chartName,omitempty"`
+	ChartUrl     *string        `db:"chart_url" json:"chartUrl,omitempty"`
+	ChartVersion *string        `db:"chart_version" json:"chartVersion,omitempty"`
+
+	// awful but relevant: the following must be defined after the ChartType, because somehow order matters
+	// for pgx at collecting the subrows (relevant at getting application + list of its versions with these
+	// array aggregations) – long term it should probably be refactored because this is such a pitfall
```
kosmoz, line 29: added link to "documentation" :sweat_smile: 
```suggestion
	// array aggregations) – long term it should probably be refactored because this is such a pitfall
	// https://github.com/jackc/pgx/issues/1585#issuecomment-1528810634
```

PR #236, opened 2025-01-08
feat(ui): make deployment target selectable in uptime chart

christophenne:


frontend/cloud-ui/src/app/components/charts/uptime/chart-uptime.component.ts
```diff
@@ -1,87 +1,104 @@
...
   private readonly metrics = inject(MetricsService);
 
+  showDropdown = false;
+  storedDeploymentTargetId?: string;
+  selectedDeploymentTarget?: DeploymentTarget;
+
+  protected readonly faEllipsis = faEllipsis;
+
   async ngOnInit() {
+    this.storedDeploymentTargetId = window.localStorage[this.LOCAL_STORAGE_KEY];
```
kosmoz, line 35: I think it would be nice to show the first deployment target if `storedDeploymentTargetId` is undefined.

Also, if there is no deployment target with that ID, should probably unset the localStorage value (and also show the first one)

PR #239, opened 2025-01-08
feat: rate limit agent and auth endpoints

christophenne:


internal/handlers/agent.go
```diff
@@ -255,3 +263,14 @@ func getVerifiedDeploymentTarget(
 		return &deploymentTarget.DeploymentTarget, nil
 	}
 }
+
+var agentConnectPerTargetIdRateLimiter = httprate.NewRateLimiter(5, time.Minute)
+var agentLoginPerTargetIdRateLimiter = httprate.NewRateLimiter(5, time.Minute)
+
+var rateLimitPerAgent = httprate.Limit(
+	2*12, // as long as we have 5 sec interval: 12 resources, 12 status requests
```
kosmoz: I think I would set it a bit higher, like `2*13` or `2*14` to account for possible irregularities caused by restarts or time weirdness (idk :sweat_smile:)

PR #242, opened 2025-01-08
feat: kubernetes agent first version

kosmoz:


internal/resources/embedded.go
```diff
@@ -10,3 +12,13 @@ var embeddedFs embed.FS
 func Get(filename string) ([]byte, error) {
 	return embeddedFs.ReadFile(filename)
 }
+
+func GetTemplate(filename string) (*template.Template, error) {
```
christophenne, line 16: `filename` param is not used but I guess that this part will change anyway somehow in the follow up PR, so nothing to do here.

cmd/agent/docker/main.go
```diff
@@ -0,0 +1,57 @@
...
+		signal.Notify(sigint, syscall.SIGTERM, syscall.SIGINT)
+		<-sigint
+		logger.Info("received termination signal")
+		cancel()
+	}()
+	tick := time.Tick(interval)
+loop:
+	for ctx.Err() == nil {
+		select {
+		case <-tick:
```
christophenne, line 36: Is there a reason why the waiting period moved up (and therefore it takes 5 seconds for the agent to do something after starting)? It might not make a big difference as long as we have the 5 seconds interval, but it might once we increase the interval?

internal/handlers/agent.go
```diff
@@ -122,36 +117,64 @@ func agentLogin(w http.ResponseWriter, r *http.Request) {
...
+			w.Header().Add("X-Resource-Correlation-ID", deployment.ID)
+			respose.Deployment = &api.KubernetesAgentDeployment{
+				RevisionID:  deployment.ID, // TODO: Update to use DeploymentRevision.ID once implemented
+				ReleaseName: *deployment.ReleaseName,
+				ChartUrl:    *appVersion.ChartUrl,
+			}
+			if *appVersion.ChartType == types.HelmChartTypeRepository {
+				respose.Deployment.ChartName = *appVersion.ChartName
+				respose.Deployment.ChartVersion = *appVersion.ChartVersion
+			}
```
christophenne, line 175: Is version also not needed when chart type is `oci`? Because then we also need to change the UI forms in that regard.

Makefile
```diff
@@ -44,17 +44,32 @@ frontend-prod: node_modules
...
-	docker build -f Dockerfile.agent --tag ghcr.io/glasskube/cloud-agent --build-arg VERSION=$(VERSION) --build-arg COMMIT=$(COMMIT) --network host .
+.PHONY: docker-build-docker-agent
+docker-build-docker-agent:
+	docker build -f Dockerfile.docker-agent --tag ghcr.io/glasskube/cloud/docker-agent --build-arg VERSION=$(VERSION) --build-arg COMMIT=$(COMMIT) --network host .
+
+.PHONY: docker-build-kubernetes-agent
+docker-build-kubernetes-agent:
+	docker build -f Dockerfile.kubernetes-agent --tag ghcr.io/glasskube/cloud/kubernetes-agent --build-arg VERSION=$(VERSION) --build-arg COMMIT=$(COMMIT) --network host .
+
+.PHONY: docker-build-kubernetes-agent
```
christophenne:
```suggestion
.PHONY: docker-build
```
? 

not a makefile expert though

internal/agentclient/client.go
```diff
@@ -0,0 +1,171 @@
...
+		} else {
+			return getCorrelationID(resp), &result, nil
+		}
+	}
+}
+
+func (c *Client) Status(ctx context.Context, correlationID, status string, error any) error {
+	body := map[string]string{
+		"status": status,
+		"error":  fmt.Sprint(error),
```
christophenne: I think `error` should only be added to the object if not nil, otherwise we receive the `"nil"` string in the status table.

christophenne: On the other hand that's maybe something for the status reporting ticket, so it can also be left like that for now.

internal/resources/embedded.go
kosmoz: yeah, I already fixed it in the follow up :-)

cmd/agent/docker/main.go
kosmoz: Yes, it's because I use `continue` a lot to break out of the loop in case of an error. Without this at the top, errors would ignore the interval.

internal/handlers/agent.go
kosmoz: No, it is needed you're right

internal/resources/embedded.go
kosmoz: nvm, I pushed it to this branch

internal/agentclient/client.go
kosmoz: fixed

internal/handlers/agent.go
```diff
@@ -165,13 +165,13 @@ func agentResourcesHandler(w http.ResponseWriter, r *http.Request) {
 			// TODO: parse values yaml and merge
 			w.Header().Add("X-Resource-Correlation-ID", deployment.ID)
 			respose.Deployment = &api.KubernetesAgentDeployment{
-				RevisionID:  deployment.ID, // TODO: Update to use DeploymentRevision.ID once implemented
-				ReleaseName: *deployment.ReleaseName,
-				ChartUrl:    *appVersion.ChartUrl,
+				RevisionID:   deployment.ID, // TODO: Update to use DeploymentRevision.ID once implemented
+				ReleaseName:  *deployment.ReleaseName,
+				ChartUrl:     *appVersion.ChartUrl,
+				ChartVersion: *appVersion.ChartName,
```
christophenne:
```suggestion
				ChartVersion: *appVersion.ChartVersion,
```

kosmoz: :facepalm:

PR #249, opened 2025-01-09
fix(ui): error handling

christophenne:


frontend/cloud-ui/src/app/password-reset/password-reset.component.ts
```diff
@@ -25,23 +26,23 @@ export class PasswordResetComponent {
...
+        location.assign(`/login?email=${encodeURIComponent(this.email)}`);
       } catch (e) {
-        if (e instanceof HttpErrorResponse && e.status < 500 && typeof e.error === 'string') {
-          this.errorMessage = e.error;
-        } else {
-          this.errorMessage = 'something went wrong';
-        }
-        console.error(e);
+        this.errorMessage = getFormDisplayedError(e);
+        // TODO maybe check for 429 again and disable the button for some time? (but how long then?)
```
kosmoz: I don't know if it's possible to inspect the response headers via the HttpErrorResponse, but the rate limiter should set a `Retry-After` header:

https://github.com/go-chi/httprate/blob/9d627fb3c8a009b20ae198abe84b3a1d10a22a94/limiter.go#L100

christophenne: The `Retry-After` header is being sent, but somehow not available in the `HttpErrorResponse`. 

Also the comment is on the wrong place, because this endpoint isn't rate limited, but rather the request one. And there, with the changes of this PR, we disable the button once the call succeeded. 

For all the other places, for now I would maybe simply change the toast message to "try again in a minute"?

kosmoz: Fine by me, but FWIW I was able to get the header using `err.headers.get('Retry-After')`

christophenne: Ok it wasn't displayed when doing `console.log`. Then I'll use the header to display the precise time.

kosmoz: I know what you mean, the problem there is that Angular inizializes the map lazily, so console.log(err.headers) only shows empty headers .....

PR #258, opened 2025-01-12
feat: organization branding

pmig:


internal/db/organization_branding.go
```diff
@@ -0,0 +1,108 @@
...
+		"SELECT"+organizationBrandingOutputExpr+
+			"FROM OrganizationBranding b "+
+			"WHERE b.organization_id = @organizationId",
+		pgx.NamedArgs{"organizationId": organizationId})
+	if err != nil {
+		return nil, fmt.Errorf("failed to query OrganizationBranding: %w", err)
+	}
+	result, err := pgx.CollectExactlyOneRow(rows, pgx.RowToStructByName[types.OrganizationBranding])
+	if errors.Is(err, pgx.ErrNoRows) {
+		return &types.OrganizationBranding{}, nil // return empty OrganizationBranding if not found
```
christophenne: but why? wouldn't it be much more obvious to send a 404 instead of an object with nil values?

internal/migrations/sql/3_organization_branding.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS OrganizationBranding
```
christophenne: version of the migration would now be 5

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,123 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		if _, err = fmt.Fprintln(w, err); err != nil {
+			log.Error("failed to write error to response", zap.Error(err))
+		}
+	} else if err = json.NewEncoder(w).Encode(organizationBranding); err != nil {
+		log.Error("failed to encode json", zap.Error(err))
+	}
+}
+
+func parseRequest(r *http.Request) (types.OrganizationBranding, error) {
```
christophenne: I would prefer it to return a pointer, then nil can be used in error cases, and in success case it doesn't copy the struct (which also includes the image).

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := r.ParseMultipartForm(102400); err != nil {
+		return types.OrganizationBranding{}, errors.New("could not parse form")
+	}
+	organizationBranding := types.OrganizationBranding{
+		Title:       r.Form.Get("title"),
+		Description: r.Form.Get("description"),
+	}
+
+	organizationBranding.ID = chi.URLParam(r, "organizationBrandingId")
```
christophenne: better use `r.PathValue` to make it chi-independent

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.ts
```diff
@@ -0,0 +1,80 @@
...
+      const id = (await lastValueFrom(this.organizationBranding$))?.id;
+
+      if (this.form.value.logo) {
+        formData.set('title', this.form.value.title!!);
+        formData.set('description', this.form.value.description!!);
+        formData.set('logo', this.form.value.logo as File);
+      }
+
+      if (id) {
+        formData.set('id', id);
```
christophenne, line 65: missing some error handling (try-catch copy pasta from other components)

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.ts
```diff
@@ -0,0 +1,80 @@
...
+        logo: this.base64ToBlob(branding.logo!!, branding.logoContentType),
+      });
+    });
+  }
+
+  async save() {
+    if (this.form.valid) {
+      const formData = new FormData();
+
+      const id = (await lastValueFrom(this.organizationBranding$))?.id;
```
christophenne: I think it should be sufficient to get the object once when loading the page, and then storing it in the component, instead of re-requesting it here?

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.ts
```diff
@@ -0,0 +1,80 @@
...
+      this.form.patchValue({
+        title: branding.title,
+        description: branding.description,
+        logo: this.base64ToBlob(branding.logo!!, branding.logoContentType),
+      });
+    });
+  }
+
+  async save() {
+    if (this.form.valid) {
```
christophenne, line 56: missing form validation feedback on the template + `markAllAsTouched` here

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
```diff
@@ -0,0 +1,88 @@
...
+
+    <form [formGroup]="form" (ngSubmit)="save()">
+      <div class="grid gap-4 mb-4 lg:grid-cols-2 sm:mb-6 mt-6">
+        <div class="space-y-4">
+          <h2 class="text-xl font-bold dark:text-white">Company Logo</h2>
+          <div>
+            <label class="block mb-2 text-sm font-medium text-gray-900 dark:text-white" for="file_input"> Logo </label>
+            <div>
+              @if (this.form.value.logo !== null) {
+                <img class="h-8 mb-4" [src]="getLogoString()" alt="Logo" />
```
christophenne: there is some error in the console in relation to the `src`

Screenshot from 2025-01-13 14-25-17

christophenne: also, at every form change (e.g. typing a letter into a text field), this error appears again and it also seems to try to load some blob from the backend. haven't quite figured this out yet

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
```diff
@@ -0,0 +1,88 @@
...
+          <h2 class="text-xl font-bold dark:text-white mt-8">Welcome Page</h2>
+
+          <div>
+            <label for="title" class="block mb-2 text-sm font-medium text-gray-900 dark:text-white">Title</label>
+            <input
+              formControlName="title"
+              type="text"
+              name="title"
+              id="title"
+              value="The 4th Digital Tranformation"
```
christophenne: can be removed

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  onLogoChange(event: any) {
+    const file = (event.target as HTMLInputElement).files?.[0];
+    this.form.patchValue({logo: file ?? null});
+  }
+
+  deleteLogo() {
+    this.form.patchValue({logo: null});
```
christophenne, line 89: logo deletion doesn't work (on the PUT request, only the id property is set because logo is null, so the backend will never know to overwrite the logo)

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
```diff
@@ -0,0 +1,88 @@
...
+              class="bg-gray-50 border border-gray-300 text-sm text-gray-900 rounded-lg focus:ring-primary-600 focus:border-primary-600 block w-full p-2.5 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500"
+              placeholder="Ex. Meeting with Adam"
+              required="" />
```
christophenne: rather make the form control required with an angular validator (also for the text area)

internal/db/organization_branding.go
pmig: If the backend returns 404 we need custom error handling in the client.

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
christophenne: I kind of got it. At every change detection cycle, the `getLogoString()` was evaluated again. Not totally sure why that leads to the expression whatever error.
Anyway, fixed it by using an observable that creates the logo string only whenever the logo form value changes.

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,123 @@
...
+	internalctx "github.com/glasskube/cloud/internal/context"
+	"github.com/glasskube/cloud/internal/db"
+	"github.com/glasskube/cloud/internal/types"
+	"github.com/go-chi/chi/v5"
+	"go.uber.org/zap"
+)
+
+func OrganizationBrandingRouter(r chi.Router) {
+	r.Get("/", getOrganizationBranding)
+	r.Post("/", createOrganizationBranding)
```
christophenne: there should also be a restriction that only the vendor is allowed to POST/PUT

internal/migrations/sql/5_organization_branding.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS OrganizationBranding
+(
+  id                         UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+  created_at                 TIMESTAMP        DEFAULT current_timestamp,
+  organization_id            UUID UNIQUE NOT NULL REFERENCES Organization (id) ON DELETE CASCADE,
+  updated_at                 TIMESTAMP        DEFAULT current_timestamp,
+  updated_by_user_account_id UUID        REFERENCES UserAccount (id) ON DELETE SET NULL,
+  title                      TEXT,
```
christophenne, line 8: @pmig I just saw now that all the properties are nullable, and therefore would also not be required on the client (what I previously assumed because the html input items had the required attribute). Is that what you had in mind or would you prefer everything required? I think it makes sense that nothing is required. 

(Logo is not required right now and is also removable already)

pmig: Yes everything should be optional / null able / empty string

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/glasskube/cloud/internal/db"
+	"github.com/glasskube/cloud/internal/types"
+	"github.com/go-chi/chi/v5"
+	"go.uber.org/zap"
+)
+
+func OrganizationBrandingRouter(r chi.Router) {
+	r.Get("/", getOrganizationBranding)
+	r.With(requireUserRoleVendor).Group(func(r chi.Router) {
+		r.Post("/", createOrganizationBranding)
```
christophenne, line 26: on another note, we could have simply done only one PUT request, because PUT means both create or update, and by reading the path variable (branding id) we could decide whether to do a database insert or update. Because the two endpoints are the same code except for the db call.

christophenne: + as long as a user can only be in one organization (which will of course change), there is no need for the branding ID in the path, it can only ever be the branding object of the current organization (which is in the token)

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		if _, err = fmt.Fprintln(w, err); err != nil {
+			log.Error("failed to write error to response", zap.Error(err))
+		}
+	} else if err = json.NewEncoder(w).Encode(organizationBranding); err != nil {
+		log.Error("failed to encode json", zap.Error(err))
+	}
+}
+
+func parseRequest(r *http.Request) (*types.OrganizationBranding, error) {
```
kosmoz: since this function is in the general "handler" package, I think it would be good to give it a more descriptive name, such as getOrganizationBrandingFromRequest or something along those lines

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+	} else {
+		organizationBranding.Logo = data
+		organizationBranding.LogoFileName = &head.Filename
+		organizationBranding.LogoContentType = util.PtrTo(head.Header.Get("Content-Type"))
+	}
+
+	return &organizationBranding, nil
+}
+
+func setMetadata(t *types.OrganizationBranding, ctx context.Context) error {
```
kosmoz: same as above, and also, the ctx argument is usually the first

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+	} else if err = db.CreateOrganizationBranding(r.Context(), organizationBranding); err != nil {
+		log.Warn("could not create organizationBranding", zap.Error(err))
+		sentry.GetHubFromContext(r.Context()).CaptureException(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		if _, err = fmt.Fprintln(w, err); err != nil {
+			log.Error("failed to write error to response", zap.Error(err))
+		}
+	} else if err = json.NewEncoder(w).Encode(organizationBranding); err != nil {
+		log.Error("failed to encode json", zap.Error(err))
+	}
```
kosmoz: new handlers should use `RespondJSON(w, ...)`

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+	} else if err := setMetadata(organizationBranding, ctx); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	} else if err = db.UpdateOrganizationBranding(r.Context(), organizationBranding); err != nil {
+		log.Warn("could not create organizationBranding", zap.Error(err))
+		sentry.GetHubFromContext(r.Context()).CaptureException(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		if _, err = fmt.Fprintln(w, err); err != nil {
+			log.Error("failed to write error to response", zap.Error(err))
+		}
+	} else if err = json.NewEncoder(w).Encode(organizationBranding); err != nil {
```
kosmoz: new handlers should use `RespondJSON(w, ...)`

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	} else if err := setMetadata(organizationBranding, ctx); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	} else if err = db.UpdateOrganizationBranding(r.Context(), organizationBranding); err != nil {
+		log.Warn("could not create organizationBranding", zap.Error(err))
+		sentry.GetHubFromContext(r.Context()).CaptureException(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		if _, err = fmt.Fprintln(w, err); err != nil {
+			log.Error("failed to write error to response", zap.Error(err))
+		}
```
kosmoz:
```suggestion
		http.Error(w, err.Error(), http.StatusInternalServerError)
```

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	} else if err := setMetadata(organizationBranding, ctx); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	} else if err = db.CreateOrganizationBranding(r.Context(), organizationBranding); err != nil {
+		log.Warn("could not create organizationBranding", zap.Error(err))
+		sentry.GetHubFromContext(r.Context()).CaptureException(err)
+		w.WriteHeader(http.StatusInternalServerError)
+		if _, err = fmt.Fprintln(w, err); err != nil {
+			log.Error("failed to write error to response", zap.Error(err))
+		}
```
kosmoz:
```suggestion
		http.Error(w, err.Error(), http.StatusInternalServerError)
```

internal/handlers/organization_branding.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func getOrganizationBranding(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	if orgID, err := auth.CurrentOrgId(ctx); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	} else if organizationBranding, err :=
+		db.GetOrganizationBranding(r.Context(), orgID); errors.Is(err, apierrors.ErrNotFound) {
+		w.WriteHeader(http.StatusNotFound)
```
kosmoz:
```suggestion
		http.NotFound(w, r)
```

internal/migrations/sql/5_organization_branding.up.sql
```diff
@@ -0,0 +1,15 @@
...
+  description                TEXT,
+  logo                       BYTEA,
+  logo_file_name             TEXT,
```
kosmoz, line 11: I think the logo file name is never used. why do we need it?

frontend/cloud-ui/src/app/services/markdown-options.factory.ts
```diff
@@ -0,0 +1,25 @@
+import {MarkedOptions, MarkedRenderer} from 'ngx-markdown';
+
+export function markedOptionsFactory(): MarkedOptions {
+  const renderer = new MarkedRenderer();
```
kosmoz, line 4: Hyper links are almost impossible to spot. I think there should be extra styling for those

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.ts
```diff
@@ -0,0 +1,91 @@
...
+      } catch (e) {
+        const msg = getFormDisplayedError(e);
+        if (msg) {
+          this.toast.error(msg);
+        }
+      }
+    }
+  }
+
+  onLogoChange(event: any) {
```
kosmoz:
```suggestion
  onLogoChange(event: Event) {
```

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
```diff
@@ -0,0 +1,86 @@
...
+            >
+            <div
+              class="mb-4 w-full bg-gray-100 rounded-lg border border-gray-200 dark:bg-gray-600 dark:border-gray-600">
+              <div class="py-2 px-4 bg-gray-50 rounded-b-lg dark:bg-gray-700">
+                <textarea
+                  formControlName="description"
+                  id="description"
+                  rows="8"
+                  class="block px-0 w-full text-sm text-gray-800 bg-gray-50 border-0 dark:bg-gray-700 focus:ring-0 dark:text-white dark:placeholder-gray-400"
+                  placeholder="Write event description..."></textarea>
```
kosmoz: event description?

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
```diff
@@ -0,0 +1,86 @@
...
+    <h1 class="mb-4 text-3xl font-extrabold leading-none tracking-tight text-gray-900 md:text-4xl dark:text-white">
+      Branding
+    </h1>
+    <p class="text-gray-500 dark:text-gray-400">
+      Customize your organization's branding by uploading a logo and setting a title and description for your
+      organization. Your customers will see this information when they access their customer Portal.
+    </p>
+
+    <form [formGroup]="form" (ngSubmit)="save()">
+      <div class="grid gap-4 mb-4 lg:grid-cols-2 sm:mb-6 mt-6">
```
kosmoz:
```suggestion
  <div class="mx-auto max-w-screen-md px-4 lg:px-12">
    <h1 class="mb-4 text-3xl font-extrabold leading-none tracking-tight text-gray-900 md:text-4xl dark:text-white">
      Branding
    </h1>
    <p class="text-gray-500 dark:text-gray-400">
      Customize your organization's branding by uploading a logo and setting a title and description for your
      organization. Your customers will see this information when they access their customer Portal.
    </p>

    <form [formGroup]="form" (ngSubmit)="save()">
      <div class="grid gap-4 mb-4 sm:mb-6 mt-6">
```
I think it looks better if you shift the layout to the center and make the form full width

frontend/cloud-ui/src/app/components/nav-bar/nav-bar.component.ts
```diff
@@ -34,12 +39,29 @@ export class NavBarComponent implements OnInit {
...
       console.error(e);
     }
   }
 
+  private async initBranding() {
+    if (this.auth.hasRole('customer')) {
+      try {
+        const branding = await lastValueFrom(this.organizationBranding.get());
+        if (branding.logo) {
+          this.logoUrl = URL.createObjectURL(base64ToBlob(branding.logo, branding.logoContentType));
```
kosmoz: Wouldn't it be easier to create a base64 data url directly like `data:${branding.logoContentType};base64,${branding.logo}` and remove the strange base64toBlob function?

frontend/cloud-ui/src/app/services/organization-branding.service.ts
```diff
@@ -0,0 +1,40 @@
+import {HttpClient} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {Observable, of, tap} from 'rxjs';
+import {OrganizationBranding, OrganizationBrandingWithAuthor} from '../types/organization-branding';
+import {CrudService} from './interfaces';
+import {DeploymentTarget} from '../types/deployment-target';
```
kosmoz: remove unused imports

frontend/cloud-ui/src/app/types/organization-branding.ts
```diff
@@ -0,0 +1,14 @@
...
+  description?: string;
+  logo?: string;
+  logoFileName?: string;
+  logoContentType?: string;
+}
+
+export interface OrganizationBrandingWithAuthor extends OrganizationBranding {
+  updatedAt: string;
+  updatedBy: string;
+}
```
kosmoz: unused

frontend/cloud-ui/src/app/components/home/home.component.ts
```diff
@@ -0,0 +1,26 @@
...
+@Component({
+  selector: 'app-home',
+  imports: [AsyncPipe, MarkdownPipe],
+  providers: [
+    provideMarkdown({
+      markedOptions: {
+        provide: MARKED_OPTIONS,
+        useFactory: markedOptionsFactory,
+      },
+    }),
```
kosmoz: this should be moved to app.config.ts

internal/routing/routing.go
```diff
@@ -54,6 +54,7 @@ func ApiRouter(logger *zap.Logger, db *pgxpool.Pool, mailer mail.Mailer) http.Ha
 			r.Route("/deployment-targets", handlers.DeploymentTargetsRouter)
 			r.Route("/user-accounts", handlers.UserAccountsRouter)
 			r.Route("/settings", handlers.SettingsRouter)
+			r.Route("/organization-branding", handlers.OrganizationBrandingRouter)
```
kosmoz: maybe change to `organization/branding`? I think this would fit better if we add more "organization" endpoints in the future

internal/migrations/sql/5_organization_branding.up.sql
christophenne: Noticed that as well, but I thought it doesn't hurt :man_shrugging: Can also remove it.

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
christophenne: @pmig your call

internal/migrations/sql/5_organization_branding.up.sql
pmig: Let's please keep it, if we want to move the files to s3 we can give them their original file names. else all the files just have an id as the file name.

kosmoz: how would you handle duplicates then?

pmig: we probably still need to put them into an id folder or something, but at least have this information =)

frontend/cloud-ui/src/app/organization-branding/organization-branding.component.html
pmig: yes, actually looks better

PR #270, opened 2025-01-14
feat: add deployment status and deployment polling

pmig:


frontend/cloud-ui/src/app/services/deployment-targets.service.ts
```diff
@@ -23,6 +23,13 @@ export class DeploymentTargetsService implements CrudService<DeploymentTarget> {
     return this.cache.get();
   }
 
+  poll(): Observable<DeploymentTarget[]> {
+    return interval(5000).pipe(
+      startWith(0),
```
kosmoz:
```suggestion
    return timer(0, 5000).pipe(
```

frontend/cloud-ui/src/app/services/deployment.service.ts
```diff
@@ -17,4 +17,11 @@ export class DeploymentService {
   getStatuses(depl: Deployment): Observable<DeploymentStatus[]> {
     return this.httpClient.get<DeploymentStatus[]>(`${this.baseUrl}/${depl.id}/status`);
   }
+
+  pollStatuses(depl: Deployment): Observable<DeploymentStatus[]> {
+    return interval(5000).pipe(
+      startWith(0),
```
kosmoz:
```suggestion
    return timer(0, 5000).pipe(
```

PR #276, opened 2025-01-15
feat: kubernetes agent self updates

kosmoz:


internal/handlers/agent.go
```diff
@@ -196,11 +181,11 @@ func agentResourcesHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// not in a TX because insertion should not be rolled back when the cleanup fails
-	if err := db.CreateDeploymentTargetStatus(ctx, deploymentTarget, statusMessage); err != nil {
+	if err := db.CreateDeploymentTargetStatus(ctx, &deploymentTarget.DeploymentTarget, statusMessage); err != nil {
```
christophenne, line 180: We could additionally store the agent version with the deployment target status.

christophenne: Maybe it could help in troubleshooting situations? But probably not high prio

internal/migrations/sql/5_agent_version.up.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE AgentVersion (
```
christophenne: version of the migration has to be updated before merging

internal/db/deployment_targets.go
```diff
@@ -169,16 +178,27 @@ func UpdateDeploymentTarget(ctx context.Context, dt *types.DeploymentTargetWithC
...
 
 	db := internalctx.GetDb(ctx)
-	args := pgx.NamedArgs{"id": dt.ID, "name": dt.Name, "orgId": orgId, "lat": nil, "lon": nil}
+	args := pgx.NamedArgs{
+		"id":             dt.ID,
+		"name":           dt.Name,
+		"orgId":          orgId,
+		"lat":            nil,
+		"lon":            nil,
+		"agentVersionId": dt.AgentVersionID,
```
christophenne, line 187: minor, but the generate scripts are failing because they don't set an agent version yet

internal/handlers/agent.go
```diff
@@ -267,23 +252,57 @@ func queryAuthDeploymentTargetCtxMiddleware(next http.Handler) http.Handler {
...
 		} else if deploymentTarget, err :=
 			db.GetDeploymentTarget(ctx, targetId, &orgId); errors.Is(err, apierrors.ErrNotFound) {
 			w.WriteHeader(http.StatusNotFound)
 		} else if err != nil {
-			internalctx.GetLogger(r.Context()).Error("failed to get DeploymentTarget", zap.Error(err))
+			log.Error("failed to get DeploymentTarget", zap.Error(err))
 			w.WriteHeader(http.StatusInternalServerError)
 		} else {
-			ctx = internalctx.WithDeploymentTarget(ctx, &deploymentTarget.DeploymentTarget)
+			if ua := r.UserAgent(); strings.HasPrefix(ua, "GlasskubeAgentClient/") {
```
christophenne: should be a constant

internal/handlers/agent.go
```diff
@@ -267,23 +252,57 @@ func queryAuthDeploymentTargetCtxMiddleware(next http.Handler) http.Handler {
...
+				reportedVersionName := strings.TrimPrefix(ua, "GlasskubeAgentClient/")
+				if reportedVersion, err := db.GetAgentVersionWithName(ctx, reportedVersionName); err != nil {
+					log.Error("could not get reported agent version", zap.Error(err))
+					sentry.GetHubFromContext(ctx).CaptureException(err)
+				} else if deploymentTarget.ReportedAgentVersionID == nil ||
+					reportedVersion.ID != *deploymentTarget.ReportedAgentVersionID {
+					if err := db.UpdateDeploymentTargetReportedAgentVersionID(
```
christophenne, line 285: Also very minor and I'm unsure, but maybe this could be moved into the handler functions, such that status insertion/deletion + this reporting update is all at the same place? In the current way its a bit scattered.

internal/handlers/agent.go
kosmoz: I thought about this too but I think we can tackle this as a separate feature

internal/handlers/agent.go
kosmoz: hmm not sure, I thought it would be convenient to just do it for all the agent requests. IMO it would kind of be more scattered to do it in every individual handler function...

PR #282, opened 2025-01-15
feat(ui): hide chart name when repo type is oci

christophenne:


frontend/cloud-ui/src/app/components/onboarding-wizard/onboarding-wizard.component.ts
```diff
@@ -330,7 +330,7 @@ export class OnboardingWizardComponent implements OnInit, OnDestroy {
       return {
         name: versionFormVal.versionName!,
         chartType: versionFormVal.chartType!,
-        chartName: versionFormVal.chartName,
+        chartName: versionFormVal.chartType === 'repository' ? versionFormVal.chartName : undefined,
```
kosmoz: shouldn't the value of `chartName` be `undefined` anyways if the related form control is disabled?

christophenne: I also thought so, but it's not. I changed the form control back to nullable and now it works as intended.

frontend/cloud-ui/src/app/components/onboarding-wizard/onboarding-wizard.component.ts
```diff
@@ -326,11 +326,12 @@ export class OnboardingWizardComponent implements OnInit, OnDestroy {
         name: this.applicationForm.controls.docker.controls.versionName.value!,
       };
     } else {
+      console.log(this.applicationForm.controls.kubernetes);
```
kosmoz:
```suggestion
```

PR #283, opened 2025-01-15
feat(ui): add application version copy

pmig:


internal/handlers/agent.go
```diff
@@ -139,6 +139,7 @@ func agentResourcesHandler(w http.ResponseWriter, r *http.Request) {
 	// TODO: Consider consolidating all types into the same response format
 	if deploymentTarget.Type == types.DeploymentTypeDocker {
 		if deployment != nil && appVersion != nil {
+			// TODO patch the compose file with a generated project name (short hash from deployment id)
```
christophenne, line 142: follow up PR because it has nothing to do with this

PR #304, opened 2025-01-16
feat: add agent docker config for kubernetes agent

kosmoz:


internal/resources/embedded/agent/kubernetes/v1/manifest.yaml.tmpl
```diff
@@ -104,3 +118,15 @@ spec:
             limits:
               memory: "128Mi"
               cpu: "500m"
+          volumeMounts:
+            - name: dockerconfig
+              mountPath: /opt/config/.docker/
+            - name: cache
```
christophenne, line 124: what is this needed for?

kosmoz: If the docker config contains invalid credentials, the chart downloader tries to access the cache for some reason. This would lead to a "file does not exist" error which is not helpful. If the cache directory exists (which is probably a good thing anyways) it leads to a way more useful 401/403 error.

christophenne: ok I see :+1:

PR #322, opened 2025-01-20
fix(ui): yaml editor respects changes

christophenne:


frontend/cloud-ui/src/app/components/yaml-editor.component.ts
```diff
@@ -48,18 +40,44 @@ export class YamlEditorComponent implements OnInit, OnDestroy {
...
-  public ngOnDestroy(): void {
+  ngOnDestroy() {
     this.destroyed$.next();
     this.destroyed$.complete();
     this.view.destroy();
   }
+
+  writeValue(value: string) {
+    const tr = this.view.state.update({changes: {from: 0, to: this.view.state.doc.length, insert: value ?? ''}});
+    this.view.dispatch(tr);
```
christophenne, line 62: This is the actual fix, the rest is just refactoring the component to a control value accessor, which I think is the better concept for this use case. 

I got the idea from the angular wrapper of codemirror: https://github.com/scttcper/ngx-codemirror/blob/master/src/lib/codemirror.component.ts (although the rest is implemented completely different).

kosmoz: Thanks :slightly_smiling_face: I looked at ngx-codemirror initially, but it is based on an older version of codemirror that is no longer developed

PR #325, opened 2025-01-20
feat: generate docker compose project name 

christophenne:


internal/handlers/agent.go
```diff
@@ -200,6 +211,17 @@ func agentResourcesHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func patchProjectName(yaml map[string]any, deploymentId string) error {
+	h := sha256.New()
+	if _, err := h.Write([]byte(deploymentId)); err != nil {
+		return err
+	} else {
+		hsh := hex.EncodeToString(h.Sum(nil))
+		yaml["name"] = fmt.Sprintf("glasskube-%v", hsh[:12])
```
kosmoz: Since we're not taking the full hash here and the deployment ID is already supposedly random, do we actually gain anything here over just using the first couple of bytes of the deploymentID instead? :thinking:

api/agent.go
```diff
@@ -8,7 +8,7 @@ type AgentResource struct {
 
 type DockerAgentResource struct {
 	AgentResource
-	ComposeFile []byte `json:"composeFile"`
+	Compose map[string]any `json:"compose"`
```
kosmoz: Is this change in anticipation of some future feature which makes it necessary? because right now I think it would be simpler to re-encode the yaml on the server-side.

internal/handlers/agent.go
christophenne: Thought the same thing but I couldn't decide. I went for this because I thought it's "more random" than the UUID, but it really doesn't matter I guess.

api/agent.go
christophenne: No I did not have such a thing in mind. Will change it back.

PR #329, opened 2025-01-21
feat: add cloud-sdk

christophenne:


.github/workflows/release.yaml
```diff
@@ -23,3 +23,28 @@ jobs:
         uses: googleapis/release-please-action@7987652d64b4581673a76e33ad5e98e3dd56832f # v4.1.3
         with:
           token: ${{ secrets.GLASSKUBE_BOT_SECRET }}
+
```
christophenne: Mostly copy pasta from the theme, so I am not sure if it works. Surely needs the npm js token to be set @kosmoz / @pmig

kosmoz: please move it to a separate workflow file that runs on tag, it's more maintainable that way. 

See release-artifacts.yaml for example.

Alternatively add the publishing step to the build-sdk.yaml but with `if: ${{ startsWith(github.ref, 'refs/tags/') }}`

kosmoz: just added the npmjs token

christophenne: ok thanks

sdk/js/src/examples/check-deployment-outdated.ts
```diff
@@ -0,0 +1,8 @@
+import {CloudService} from '../client/service';
+import {clientConfig, getSomeDockerDeploymentTargetId} from './helper';
+
+const gc = new CloudService(clientConfig);
+
+const deploymentTargetId = '79138254-31c1-4e09-b511-4862d257c80d'; // await getSomeDeploymentTargetId();
```
pmig: `getSomeDockerDeploymentTargetId` unsused or on purpose?

sdk/js/src/examples/update-deployment-kubernetes.ts
```diff
@@ -0,0 +1,7 @@
+import {CloudService} from '../client/service';
+import {clientConfig, getSomeKubernetesDeploymentTargetId} from './helper';
+
+const gc = new CloudService(clientConfig);
+
+const deploymentTargetId = '755e9329-cc32-407a-aa59-d220202cf899'; // await getSomeKubernetesDeploymentTargetId(clientConfig);
```
pmig: `getSomeKubernetesDeploymentTargetId` unsused or on purpose?

sdk/js/src/client/index.ts
```diff
@@ -0,0 +1 @@
+export * from './client';
```
kosmoz, line 1: also export `./service`?

sdk/js/src/client/client.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+export type ApplicationVersionFiles = {
+  composeFile?: string;
+  baseValuesFile?: string;
+  templateFile?: string;
+};
+
+export class Client {
+  constructor(private readonly config: ClientConfig) {}
```
kosmoz: IMO we should provide a default apiBase here. I think the cleanes way to do this would be something like this:

```suggestion
export type ClientConfig = {
  apiBase: string;
  apiKey: string;
};

export type ApplicationVersionFiles = {
  composeFile?: string;
  baseValuesFile?: string;
  templateFile?: string;
};

const defaultClientConfig = {apiBase: 'https://app.glasskube.cloud/api/v1/'};

type ConditionalPartial<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;

export class Client {
  private readonly config: ClientConfig;

  constructor(config: ConditionalPartial<ClientConfig, keyof typeof defaultClientConfig>) {
    this.config = {...defaultClientConfig, ...config};
  }
```

Additionally, normalizing the apiBase to ensure that it ends with a `/` might be a good idea

.github/workflows/build-sdk.yaml
```diff
@@ -0,0 +1,32 @@
+name: Build SDK
+
+on:
+  push:
+    branches:
+      - 'main'
```
kosmoz, line 6:
```suggestion
      - 'main'
    paths:
      - sdk/**
```

sdk/js/src/client/client.ts
kosmoz: hmm I just realized this doesn't pass type checking, let me think

edit: updated suggestion

sdk/js/src/client/service.ts
```diff
@@ -0,0 +1,248 @@
...
+import {
+  Application,
+  ApplicationVersion,
+  DeploymentTarget,
+  DeploymentTargetAccessResponse,
+  DeploymentTargetScope,
+  DeploymentType,
+  HelmChartType,
+} from '../types';
+import semver from 'semver/preload';
```
kosmoz, line 11: Not sure, but I think this package is not compatible with browsers? Should the SDK be compatible with browsers?

christophenne: Good point, I'll look into that

christophenne: I tested it by using it in the frontend, and it worked for me.

kosmoz: cool, thanks for checking

sdk/js/src/client/client.ts
christophenne: I had to adapt this a bit since its also used when initializing the service.

PR #333, opened 2025-01-21
feat: api token authentication

kosmoz:


internal/authn/token/token.go
```diff
@@ -0,0 +1,44 @@
...
+
+func NewTokenExtractor(tokenFuncs ...TokenExtractorFunc) TokenExtractor {
+	return TokenExtractor(tokenFuncs)
+}
+
+func TokenFromQuery(param string) TokenExtractorFunc {
+	return func(r *http.Request) string { return r.URL.Query().Get(param) }
+}
+
+func TokenFromHeader(authenticationScheme string) TokenExtractorFunc {
```
christophenne: these names could omit the `Token` prefix since the package is already called `token`

internal/auth/authentication.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/glasskube/cloud/internal/authn/jwt"
+	"github.com/glasskube/cloud/internal/authn/token"
+)
+
+var Authentication = authn.New(
+	authn.Chain(
+		authn.Chain(
+			token.NewTokenExtractor(
+				token.TokenFromHeader("Bearer"),
+				token.TokenFromQuery("jwt"),
```
christophenne: I looked for it, but I did not find any point at where we use the token from query? Do we really need to support this?

christophenne: Oh is it because the chi authenticator said it does that? – edit: I don't think it does that. 

```
// Verifier http middleware handler will verify a JWT string from a http request.
//
// Verifier will search for a JWT token in a http request, in the order:
//  1. 'Authorization: BEARER T' request header
//  2. Cookie 'jwt' value
//
// The first JWT string that is found as a query parameter, authorization header
// or cookie header is then decoded by the `jwt-go` library and a *jwt.Token
// object is set on the request context. In the case of a signature decoding error
// the Verifier will also set the error on the request context.
//
// The Verifier always calls the next http handler in sequence, which can either
// be the generic `jwtauth.Authenticator` middleware or your own custom handler
// which checks the request context jwt token and error to prepare a custom
// http response.
func Verifier(ja *JWTAuth) func(http.Handler) http.Handler {
	return Verify(ja, TokenFromHeader, TokenFromCookie)
}
```

internal/db/access_token.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, fmt.Errorf("error querying access tokens: %w", err)
+	}
+	if result, err := pgx.CollectRows(rows, pgx.RowToStructByName[types.AccessToken]); err != nil {
+		return nil, fmt.Errorf("could not get tokens: %w", err)
+	} else {
+		return result, nil
+	}
+}
+
+func GetAccessTokenByKey(ctx context.Context, key authkey.Key) (*types.AccessTokenWithUserAccount, error) {
```
christophenne: I would rename it to something like `UseAccessToken`, because from a `Get...` I wouldn't expect it to update anything.

internal/db/access_token.go
```diff
@@ -0,0 +1,108 @@
...
+	db := internalctx.GetDb(ctx)
+	rows, err := db.Query(
+		ctx,
+		fmt.Sprintf(
+			`WITH updated AS (
+				UPDATE AccessToken
+				SET last_used_at = now()
+				WHERE key = @key AND expires_at > now()
```
christophenne: does that not fail if `expires_at is null`?

api/access_token.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/glasskube/cloud/internal/authkey"
+)
+
+type AccessToken struct {
+	ID         string     `json:"id"`
+	CreatedAt  time.Time  `json:"createdAt"`
+	ExpiresAt  *time.Time `json:"expiresAt,omitempty"`
+	LastUsedAt *time.Time `json:"lastUsedAt,omitempty"`
+	Label      string     `json:"label"`
```
christophenne: shouldn't the label be a `*string` everywhere since its optional/nullable?

frontend/cloud-ui/src/app/access-tokens/access-tokens.component.html
```diff
@@ -0,0 +1,153 @@
...
+            id="createApplicationButton"
+            class="w-full md:w-auto flex items-center justify-center py-2 px-4 text-sm font-medium text-gray-900 focus:outline-none bg-white rounded-lg border border-gray-200 hover:bg-gray-100 hover:text-primary-700 focus:z-10 focus:ring-4 focus:ring-gray-200 dark:focus:ring-gray-700 dark:bg-gray-800 dark:text-white dark:border-gray-600 dark:hover:text-white dark:hover:bg-gray-700">
+            <fa-icon [icon]="faPlus" class="text-gray-500 dark:text-gray-400 mr-2"></fa-icon>
+            Create token
+          </button>
+        </div>
+      </div>
+
+      @if (createdToken; as t) {
+        <div
```
christophenne, line 23: probably a short hint could help that this is the only time the user will see it and they need to copy it and store it safely and that kind of bla bla

christophenne: (and the classic copy button)

frontend/cloud-ui/src/app/access-tokens/access-tokens.component.html
```diff
@@ -0,0 +1,153 @@
...
+            <div class="relative w-full">
+              <div class="absolute inset-y-0 left-0 flex items-center pl-3 pointer-events-none">
+                <fa-icon [icon]="faMagnifyingGlass" class="text-gray-500 dark:text-gray-400"></fa-icon>
+              </div>
+              <input
+                [formControl]="filterForm.controls.search"
+                autotrim
+                type="text"
+                id="simple-search"
+                placeholder="Search for applications"
```
christophenne:
```suggestion
                placeholder="Search for tokens"
```

or remove the header, because its not implemented right now

frontend/cloud-ui/src/app/access-tokens/access-tokens.component.html
```diff
@@ -0,0 +1,153 @@
...
+          <label for="expiresAt" class="block mb-2 text-sm font-medium text-gray-900 dark:text-white">Expires At</label>
+          <input
+            formControlName="expiresAt"
+            type="date"
+            id="expiresAt"
+            class="bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-primary-600 focus:border-primary-600 block w-full p-2.5 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500"
+            placeholder="Label your PAT" />
+        </div>
+
+        <pre class="dark:text-white">{{ editForm.value | json }}</pre>
```
christophenne: was probably just for debugging?

internal/auth/authentication.go
kosmoz: The jwtauth middleware supports it if you enable the feature. I considered somehow using this for the connect endpoint but then decided that it would be too unrelated. It can be removed I think, but at the same time it shouldn't really hurt to have it so … :shrug:  what do you think?

frontend/cloud-ui/src/app/access-tokens/access-tokens.component.html
kosmoz: you're right, I removed it for now because it doesn't work anyways

frontend/cloud-ui/src/app/access-tokens/access-tokens.component.html
kosmoz: yes :slightly_smiling_face:

internal/auth/authentication.go
christophenne: Ok, I'm slightly more on the remove side, but it really doesn't matter too much.

internal/db/access_token.go
kosmoz: yes, good catch :+1:

internal/db/access_token.go
kosmoz: I changed the name to `GetAccessTokenByKeyUpdatingLastUsed` since that is even more explicit

internal/auth/authentication.go
kosmoz: removed :+1:

frontend/cloud-ui/src/app/access-tokens/access-tokens.component.html
kosmoz: I though I could get away with just `user-select: all` :see_no_evil:

frontend/cloud-ui/src/app/access-tokens/access-tokens.component.html
```diff
@@ -4,23 +4,7 @@
...
-                id="simple-search"
-                placeholder="Search for applications"
-                class="bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-primary-500 focus:border-primary-500 block w-full pl-10 p-2 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500" />
-            </div>
-          </form>
-        </div>
+        <div></div>
         <div
           class="w-full md:w-auto flex flex-col md:flex-row space-y-2 md:space-y-0 items-stretch md:items-center justify-end md:space-x-3 flex-shrink-0">
           <button
```
christophenne, line 10: If only a vendor can create one, a customer would always get an empty list, right? Probably makes sense to hide the menu entry and optionally a route guard.

kosmoz: Good point, that is actually a copy+paste mistake. The intention was that everybody should be able to create PATs

PR #355, opened 2025-01-24
chore: additional api request limits

christophenne:


internal/routing/routing.go
```diff
@@ -48,6 +51,9 @@ func ApiRouter(logger *zap.Logger, db *pgxpool.Pool, mailer mail.Mailer) http.Ha
 			r.Use(
 				middleware.SentryUser,
 				auth.Authentication.Middleware,
+				httprate.Limit(5, 1*time.Second, httprate.WithKeyFuncs(middleware.RateLimitCurrentUserIdKeyFunc)),
```
christophenne, line 55: actually unsure whether these should be by user or by token? If somebody does a "regular" amount of polling via SDK, they might run into rate limits when using the UI as well. 

@kosmoz / @pmig what do you think?

kosmoz: good question… per token sounds wrong, as it would open the platform for abuse since people can just generate more and more tokens. on the other hand your concern about having it per user is also valid. 

some things that might work for now:
- make it per user but only apply rate limit to requests with API tokens not with JWT
- make it per user but use a different key func for API token and JWT

christophenne: Implemented it similarly to what we said, please check.

kosmoz: Looks good, my only criticism would be that the name `RawToken` (at least to me) implies that it was not parsed yet, but in this case it has already been parsed. Maybe renaming it to `Token` or `ParsedToken` would be better?

The agent client has something similar here: 

https://github.com/glasskube/cloud/blob/49fd3742557baee91ff73aeb9fbd55a7ef1371bf/internal/agentclient/client.go#L33-L34

christophenne: Yep, will do.

PR #423, opened 2025-02-05
fix(backend): avoid panic on empty compose file

christophenne:
not sure about this though, maybe the `ParsedXXXFile()` functions should always return a non-nil map when there is no error @kosmoz ?


kosmoz:
> not sure about this though, maybe the `ParsedXXXFile()` functions should always return a non-nil map when there is no error @kosmoz ?

I'm not convinced tbh, maybe it makes sense to allow differentiating between "no compose file" and "empty compose file"? OTOH, should an application version with an empty compose file even be allowed at all?

kosmoz (approved):

PR #424, opened 2025-02-05
feat: support environment for docker compose

christophenne:
dependent PRs:
* https://github.com/glasskube/distr-create-version-action/pull/20
* https://github.com/glasskube/distr.sh/pull/26

### DIFF
--- a/frontend/ui/src/app/components/yaml-editor.component.ts
+++ b/frontend/ui/src/app/components/yaml-editor.component.ts
@@ -71,7 +71,10 @@ export class YamlEditorComponent implements OnInit, OnDestroy, ControlValueAcces
   }
 
   setDisabledState(isDisabled: boolean) {
-    // TODO don't know how to do this yet, but we probably also don't want a sentry error for that
+    // TODO implement
+    if (isDisabled) {
+      console.error('setDisabledState not implemented yet');
+    }
   }
 
   private onChange = (_: any) => {};


frontend/ui/src/app/components/yaml-editor.component.ts
```diff
@@ -71,10 +71,7 @@ export class YamlEditorComponent implements OnInit, OnDestroy, ControlValueAcces
   }
 
   setDisabledState(isDisabled: boolean) {
-    // TODO implement
-    if (isDisabled) {
-      console.error('setDisabledState not implemented yet');
-    }
+    // TODO don't know how to do this yet, but we probably also don't want a sentry error for that
```
kosmoz: For reference: https://codemirror.net/examples/readonly/ Looks like it's notthat easy to change dynamically?

QUestion: does console.error really trigger a sentry event?

frontend/ui/src/app/components/installation-wizard/installation-wizard.component.html
```diff
@@ -266,6 +266,19 @@ <h3 class="font-normal text-gray-900 dark:text-white">Deploy</h3>
                       </app-yaml-editor>
                     </div>
                   }
+                  @if (deployForm.controls.envFileData.enabled) {
+                    <div class="col-span-2">
+                      <label for="env-file" class="block mb-2 text-sm font-medium text-gray-900 dark:text-white">
+                        Environment values
+                      </label>
+
+                      <app-yaml-editor
```
kosmoz, line 275: Using YAML syntax highlighting here might be confusing since some incorrect syntax might look more "correct" in the editor than the correct syntax. At the same time, it doesn't look like changing the installed extensions on the fly is particularly easy so we would probably need a separate component without syntax highlighting for that.

So, since I think it's probably not a problem for now, we can leave as is but I would like to change this to an editor view without syntax highlighting in the future

kosmoz (commented):
Works great. Just have some questions/comments nothing to change I think

frontend/ui/src/app/components/yaml-editor.component.ts
christophenne: Good point, probably it only triggers uncaught errors, I'll revert.

frontend/ui/src/app/components/installation-wizard/installation-wizard.component.html
christophenne: True, I created an issue

kosmoz (approved):

PR #425, opened 2025-02-05
feat(docker-agent): add self-updating

kosmoz:


api/agent.go
```diff
@@ -3,23 +3,31 @@ package api
...
+	Version types.AgentVersion `json:"version"`
+}
+
+type AgentDeployment struct {
 	RevisionID string `json:"revisionId"`
 }
 
 type DockerAgentResource struct {
 	AgentResource
+	Deployment *DockerAgentDeployment `json:"deployment"`
```
christophenne, line 15: Technically that's a breaking change for existing agents, right? We should discuss how we want to deal with it.

cmd/agent/docker/main.go
```diff
@@ -53,25 +54,132 @@ loop:
...
-			var reportedStatus string
-			var reportedErr error
-			if cmdErr != nil {
-				reportedErr = errors.New(outStr)
-			} else {
-				reportedStatus = outStr
+			if agentVersionId != "" {
+				if agentVersionId != resource.Version.ID {
+					logger.Info("agent version has changed. starting self-update")
+					RunAgentSelfUpdate(ctx)
```
christophenne: Would it make sense that the agent can recover from some of these cases, and continue to serve the deployment even if the update failed?

api/agent.go
kosmoz: True, as discussed we will go forward with this change, but I updated the PR title to reflect this a bit better.

PR #427, opened 2025-02-05
feat: add showing invite URL after user creation

kosmoz:


frontend/ui/src/app/services/users.service.ts
```diff
@@ -10,6 +10,11 @@ export interface CreateUserAccountRequest {
   applicationName?: string;
 }
 
+export interface CreateUserAccountResponse {
+  id: string;
+  inviteUrl?: string;
```
christophenne: Is there really a (success-) case where the `inviteUrl` would not be given?

kosmoz: Currently, no. Initially, I wanted to only return the inviteUrl if email sending is disabled, but I later decided that it probably makes sense regardless. I will remove the optional marker

PR #452, opened 2025-02-10
fix(ui): update deployment targets list immediately

christophenne:
I moved `deploy` over to `DeploymentTargetsService`, because of the shared data they seem to be more related to each other. 

Also at this point the cache in the `DeploymentTargetsService` is almost meaningless because we refetch on every change anyway.

### DIFF
--- a/frontend/ui/src/app/deployments/deployment-targets.component.ts
+++ b/frontend/ui/src/app/deployments/deployment-targets.component.ts
@@ -33,7 +33,6 @@ import {
   firstValueFrom,
   lastValueFrom,
   map,
-  merge,
   Observable,
   of,
   Subject,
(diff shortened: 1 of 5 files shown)


frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -147,7 +148,11 @@ export class DeploymentTargetsComponent implements OnInit, AfterViewInit, OnDest
   });
 
   deployFormLoading = false;
-  readonly deploymentTargets$ = this.deploymentTargets.poll();
+  private readonly refresh = new Subject<void>();
+  readonly deploymentTargets$ = merge(
+    this.deploymentTargets.poll(),
+    this.refresh.pipe(switchMap(() => this.deploymentTargets.list()))
+  );
```
kosmoz: Is this still necessary? I think the same is done in the service already

kosmoz (commented):
I agree about the cache but at the same time it doesn't really hurt to keep it around since it is used in other places too…

kosmoz (approved):

PR #455, opened 2025-02-10
feat(backend): add db models for application licenses

kosmoz:
I also added some DB access functions that I anticipated we are going to need, but they are not tested…

### DIFF
--- a/internal/migrations/sql/11_application_license.up.sql
+++ b/internal/migrations/sql/11_application_license.up.sql
@@ -1,0 +1,33 @@
+CREATE TABLE ApplicationLicense (
+  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+  created_at TIMESTAMP DEFAULT current_timestamp,
+  name TEXT NOT NULL,
+  expires_at TIMESTAMP,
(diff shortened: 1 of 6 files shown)


internal/migrations/sql/11_application_license.up.sql
```diff
@@ -0,0 +1,33 @@
...
+);
+
+CREATE TABLE ApplicationLicense_ApplicationVersion (
+  application_license_id UUID NOT NULL REFERENCES ApplicationLicense (id),
+  application_version_id UUID NOT NULL REFERENCES ApplicationVersion (id),
+  PRIMARY KEY (application_license_id, application_version_id)
+);
+
+ALTER TABLE Deployment
+  ADD COLUMN application_license_id UUID REFERENCES ApplicationLicense (id);
```
christophenne: This just got me thinking, we recently made the decision that changing the application of a deployment makes no sense, and disabled the dropdown on the UI. But we don't yet check an application change on the backend. 
Now it would be possible to have an application version of a different application deployed, from what the license is for, which makes no sense. I think we should add a validation at the deployment endpoint soon.

PR #458, opened 2025-02-10
feat: license management feature flag

christophenne:
Unfortunately I couldn't make an enum array work with pgx, so I opted for the boolean for now. We can still revisit/refactor/migrate on the second, or latest, third occurrence of a feature flag. 

As a positive side effect, I think the manual updating of a db entry will be more straight forward for now :man_shrugging:

### DIFF
--- a/internal/svc/registry.go
+++ b/internal/svc/registry.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"os"
 	"syscall"
 
 	"github.com/glasskube/distr/internal/migrations"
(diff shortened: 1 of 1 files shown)


pmig (dismissed):
I'd rather prefer to not use a boolean but store features with an ENUM in a list / separate table.

kosmoz:
I think I found the problem, and it appears to be a bug in pgx!!

I think if you replace LoadTypes(...) and RegisterTypes(...) in the svc package with a loop around LoadType(...) and RegisterType(...) it should work!

kosmoz:
Debugging the createDBPool function … it appears that the pgTypes returned by conn.LoadTypes is just an empty slice lol

kosmoz:
I created the following related issue in the pgx repo: https://github.com/jackc/pgx/issues/2254 

For now it's probably easiest to just load each type individually.

internal/svc/registry.go
```diff
@@ -88,13 +109,18 @@ func createDBPool(ctx context.Context, log *zap.Logger) (*pgxpool.Pool, error) {
...
+			"_FEATURE"}
+		for _, typeName := range typeNames {
+			if pgType, err := conn.LoadType(ctx, typeName); err != nil {
+				fmt.Fprintf(os.Stderr, "cannot load type %s: %v\n", typeName, err)
+				return err
+			} else {
+				conn.TypeMap().RegisterType(pgType)
+			}
 		}
+		fmt.Fprintf(os.Stderr, "done loading types\n")
```
kosmoz: Is there a reason why the logger is not used here?

kosmoz (commented):
looks good, just one small question

internal/svc/registry.go
christophenne: nope was just debug output

kosmoz (approved):
looks good, but let's merge https://github.com/glasskube/distr/pull/455 fist because of the migration IDs

PR #460, opened 2025-02-11
fix(backend): change DeploymentTarget created_by reference on delete to `RESTRICT`

kosmoz:


internal/db/user_accounts.go
```diff
@@ -130,13 +130,20 @@ func UpdateUserAccountEmailVerified(ctx context.Context, userAccount *types.User
 
 func DeleteUserAccountWithID(ctx context.Context, id string) error {
 	db := internalctx.GetDb(ctx)
-	if cmd, err := db.Exec(ctx, `DELETE FROM UserAccount WHERE id = @id`, pgx.NamedArgs{"id": id}); err != nil {
-		return err
+	cmd, err := db.Exec(ctx, `DELETE FROM UserAccount WHERE id = @id`, pgx.NamedArgs{"id": id})
+	if err != nil {
+		if pgerr := (*pgconn.PgError)(nil); errors.As(err, &pgerr) && pgerr.Code == pgerrcode.ForeignKeyViolation {
```
christophenne, line 135: I don't 100% understand the `(*pgconn.PgError)(nil)` part, is it the same as having a `var pgerr *pgconn.PgError` one line above?

kosmoz: pretty much, but a simple declaration is not allowed as part of the if statement

kosmoz:
> Works fine, only noticed that the deletion confirm modal still says "Really delete test? This will also delete all deployments they manage!" – I think the second sentence should be removed now :)

good catch, fixed

PR #463, opened 2025-02-11
feat: add selecting a license for a deployment

kosmoz:


frontend/ui/src/app/components/installation-wizard/installation-wizard.component.ts
```diff
@@ -201,18 +154,18 @@ export class InstallationWizardComponent implements OnInit, OnDestroy {
...
-    this.deployForm.patchValue({
-      deploymentTargetId: this.selectedDeploymentTarget()!.id,
-    });
-
     this.deployForm.markAllAsTouched();
     if (!this.deployForm.valid || this.loading) {
       return;
     }
 
+    if (!this.deployForm.valid) {
```
christophenne: why again? it's already checked in L158

frontend/ui/src/app/deployment-form/deployment-form.component.ts
```diff
@@ -0,0 +1,304 @@
...
+        distinctUntilChanged(
+          (a, b) =>
+            a?.id === b?.id &&
+            a?.deployment?.envFileData === b?.deployment?.envFileData &&
+            a?.deployment?.valuesYaml === b?.deployment?.valuesYaml
+        )
+      ),
+    ])
+      .pipe(
+        debounceTime(5),
```
christophenne, line 238: Was it just for debugging purposes?

frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -371,7 +291,10 @@ export class DeploymentTargetsComponent implements OnInit, AfterViewInit, OnDest
     this.deployForm.markAllAsTouched();
```
christophenne, line 291: The marking as touched isn't well supported by the CVA, but I found a hack/workaround in the edit-license component: 

```
ngAfterViewInit() {
    // from https://github.com/angular/angular/issues/45089
    this.injector
      .get(NgControl)
      .control!.events.pipe(takeUntil(this.destroyed$))
      .subscribe((event) => {
        if (event instanceof TouchedChangeEvent) {
          console.log('event', event);
          if (event.touched) {
            this.editForm.markAllAsTouched();
          }
        }
      });
  }
```

internal/handlers/applications.go
```diff
@@ -108,33 +108,49 @@ func updateApplication(w http.ResponseWriter, r *http.Request) {
...
-		internalctx.GetLogger(r.Context()).Error("failed to encode to json", zap.Error(err))
+	ctx := r.Context()
+	auth := auth.Authentication.Require(ctx)
+	log := internalctx.GetLogger(ctx)
+
+	if org, err := db.GetOrganizationByID(ctx, *auth.CurrentOrgID()); err != nil {
+		log.Error("failed to get org", zap.Error(err))
+		sentry.GetHubFromContext(ctx).CaptureException(err)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+	} else if org.HasFeature(types.FeatureLicensing) && *auth.CurrentUserRole() == types.UserRoleCustomer {
```
christophenne: Might be because it's WIP, but shouldn't the customer be able to at least get the versions they have the license for? But I think the endpoint isn't used by our UI at all and could even be removed?

On a similar matter, maybe license validations for downloading the template/compose files would make sense?

frontend/ui/src/app/deployment-form/deployment-form.component.ts
kosmoz: No, because the applicationVersionId changes very shortly after each time the applicationId changes, two requests would trigger where the first one is immediately cancelled. Debouncing by 5ms is not noticable (at leat I don't :^)) and prevents this.

frontend/ui/src/app/deployments/deployment-targets.component.ts
kosmoz: beautiful :heart_eyes:

frontend/ui/src/app/components/installation-wizard/installation-wizard.component.ts
kosmoz: true, I missed that

internal/handlers/applications.go
kosmoz: Fair question… I thought about this as well and I came to the conclusion that it doesn't make a lot of sense to give the user a combined and deduplicated list of all versions from their licenses. alternatively I though of adding an applicationLicenseId query param that is required for users with role customer but at the same time the same can be achieved by just getting the license. So, for now I just return this error code instead of "leaking" versions the customer should not be able to see.

I didn't realize the UI doesn't use this endpoint at all. It appears that the SDK doesn't integrate it as well? maybe we can remove it after all and always require getting the versions from the application or license endpoint

frontend/ui/src/app/deployment-form/deployment-form.component.ts
christophenne: Ah okay

internal/handlers/applications.go
christophenne: Yes, I would remove the endpoint.

kosmoz: removed

frontend/ui/src/app/types/license.ts
```diff
@@ -0,0 +1,8 @@
+import {ApplicationVersion, BaseModel} from '@glasskube/distr-sdk';
+
+export interface License extends BaseModel {
```
christophenne: This is unused

frontend/ui/src/app/deployment-form/deployment-form.component.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+  }
+
+  writeValue(obj: DeploymentFormValue | null | undefined): void {
+    if (obj) {
+      this.deployForm.patchValue(obj);
+    } else {
+      this.deployForm.reset();
+    }
+    console.log('writeValue', obj, this.deployForm.value);
```
christophenne:
```suggestion
```

PR #476, opened 2025-02-13
feat: license management

christophenne:


kosmoz (commented):
I know this PR is still WIP, but I noticed two small things:
- the icon before the "License Mgmt" nav link still has the "disabled" appearance
- there is a build warning:
```
▲ [WARNING] NG8107: The left side of this optional chain operation does not include 'null' or 'undefined' in its type, therefore the '?.' operator can be replaced with the '.' operator. [plugin angular-compiler]

    frontend/ui/src/app/licenses/licenses.component.html:74:53:
      74 │                           ((license.versions || [])?.length === 0
         ╵                                                      ~~~~~~
```

kosmoz:
Another thing: I noticed that a license with no registry auth is saved with all related fields set to an empty string, but I think it should be using NULL instead. Either the frontend should not send those fields in case they are empty, or the backend should normalize the model before storing it in the DB

frontend/ui/src/app/licenses/edit-license.component.ts
```diff
@@ -0,0 +1,267 @@
...
+  private readonly destroyed$ = new Subject<void>();
+  private readonly applicationsService = inject(ApplicationsService);
+  private readonly usersService = inject(UsersService);
+  applications$ = this.applicationsService.list();
+  customers$ = this.usersService.getUsers().pipe(
+    map((accounts) => accounts.filter((a) => a.userRole === 'customer')),
+    first()
+  );
+
+  private fb = inject(FormBuilder);
```
kosmoz, line 64: The form builder is only used for one field here, is this necessary? I think it would be better to use it for all fields or not at all

frontend/ui/src/app/licenses/edit-license.component.ts
```diff
@@ -0,0 +1,267 @@
...
+          return {
+            required: true,
+          };
+        },
+      }
+    ),
+  });
+  editFormLoading = false;
+  license: WritableSignal<ApplicationLicense | undefined> = signal(undefined);
+  selectedApplication: WritableSignal<Application | undefined> = signal(undefined);
```
kosmoz: slightly more concise
```suggestion
  readonly license = signal<ApplicationLicense | undefined>(undefined);
  readonly selectedApplication = signal<Application | undefined>(undefined);
```

frontend/ui/src/app/licenses/edit-license.component.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  dropdownWidth: number = 0;
+
+  protected readonly faMagnifyingGlass = faMagnifyingGlass;
+  protected readonly faChevronDown = faChevronDown;
+  protected readonly faPlus = faPlus;
+  protected readonly faXmark = faXmark;
+  protected readonly faPen = faPen;
+
+  @ViewChild('dropdownTriggerButton') dropdownTriggerButton!: ElementRef;
```
kosmoz:
```suggestion
  @ViewChild('dropdownTriggerButton') dropdownTriggerButton!: ElementRef<HTMLElement>;
```

frontend/ui/src/app/licenses/edit-license.component.ts
```diff
@@ -0,0 +1,267 @@
...
+            this.editForm.markAllAsTouched();
+          }
+        }
+      });
+  }
+
+  toggleDropdown() {
+    this.dropdownOpen.update((v) => !v);
+    if (this.dropdownOpen()) {
+      this.dropdownWidth = (this.dropdownTriggerButton.nativeElement as Element).getBoundingClientRect().width;
```
kosmoz: combined with previous suggestion
```suggestion
      this.dropdownWidth = this.dropdownTriggerButton.nativeElement.getBoundingClientRect().width;
```

frontend/ui/src/app/licenses/edit-license.component.ts
```diff
@@ -0,0 +1,267 @@
...
+    this.destroyed$.next();
+    this.destroyed$.complete();
+  }
+
+  get versionsArray() {
+    return this.editForm.controls.versions as FormArray;
+  }
+
+  private onChange: (l: ApplicationLicense | undefined) => void = () => {};
+  private onTouched: () => void = () => {};
```
kosmoz, line 229: onTouched is never called. would it make sense to also call it when the value changes? if not, I think it would be better to remove this

internal/handlers/application_licenses.go
```diff
@@ -0,0 +1,278 @@
...
+				log.Warn("could not add version to license", zap.Error(err))
+				sentry.GetHubFromContext(ctx).CaptureException(err)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return err
+			}
+		}
+		return nil
+	})
+	if err == nil {
+		if completeLicense, err := db.GetApplicationLicenseByID(ctx, license.ID); err != nil {
```
kosmoz: If I understand correctly, this is needed to "backfill" the fields from relations of the license? If so, I think there is a more elegant way to do this directly with the inset using a WITH clause:
```sql
WITH inserted AS (
  -- insert statemant as normal
  RETURNING *
)
SELECT ...
  FROM inserted al -- use "inserted" instead of ApplicationLicense
    LEFT JOIN ...
  -- rest of the query
```

We do this for DeploymentTargets already. Would it make sense here? https://github.com/glasskube/distr/blob/d2cbf813dc1250fd3a37ace889613f206d13acbf/internal/db/deployment_targets.go#L153-L164

Alternatively, I think sending an incomplete response JSON in case of a failure should be avoided because it could lead to some ugly edge cases in the frontend. Instead I think it should either return an error, even though the transaction has succeeded, or roll back the transaction and return an error.

internal/handlers/application_licenses.go
```diff
@@ -0,0 +1,278 @@
...
+			return err
+		} else if err != nil {
+			log.Warn("could not update license", zap.Error(err))
+			sentry.GetHubFromContext(ctx).CaptureException(err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return err
+		}
+
+		for _, version := range license.Versions {
+			alreadyExists := false
```
kosmoz: these "alreadyExists" and "stillExists" could be simplified using something like this
```go
alreadyExists := slices.ContainsFunc(existing.Versions, func(v types.ApplicationVersion) bool { return v.ID == version.ID })
stillExists := slices.ContainsFunc(license.Versions, func(v types.ApplicationVersion) bool { return v.ID == existingVersion.ID })
```

frontend/ui/src/app/licenses/edit-license.component.ts
christophenne: I didn't find an option to build the form array, that's why I use it here. Will change the rest of it too

internal/handlers/application_licenses.go
christophenne: I looked into this, but I wasn't sure how to do it in this case. The difference to the cases where we already do that, is that the `INSERT` or `UPDATE` isn't necessarily the last statement. There might be additional `INSERT`s and `DELETE`s on the `ApplicationLicense_ApplicationVersion` table. That's why I thought it's easier to just reload whatever the outcome of the previous transaction was.

kosmoz: I see. I think it's OK not to do that here but I would still prefer to fail the entire transaction in this case instead of returning an incomplete object to the client, since then the burden would be on the client to handle the missing data somehow

christophenne: Yes, true.

kosmoz (approved):

PR #515, opened 2025-02-18
feat: add "undeploying" applications from a target

kosmoz:


cmd/agent/docker/docker_actions.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if err != nil {
+		return nil, "", errors.New(statusStr)
+	} else {
+		return agentDeploymet, statusStr, nil
+	}
+}
+
+func UninstallDockerCompose(ctx context.Context, deployment AgentDeployment) error {
+	cmd := exec.CommandContext(ctx, "docker", "compose", "--project-name", deployment.ProjectName, "down", "--volumes")
```
christophenne, line 64: More of a feature discussion than code review, but: I think destroying the volumes is very scary, can't we keep them? (I know that would also have to be done in the helm case and is probably more complicated).

christophenne: If that's not a thing at all, I think the confirmation on the frontend should be "more difficult", like having the user enter "DELETE DATA" or something like that.

frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -198,7 +208,14 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
 
   async deleteDeploymentTarget(dt: DeploymentTarget) {
     this.overlay
-      .confirm(`Really delete ${dt.name}? This action can not be undone.`)
+      .confirm(
```
christophenne: In these confirm dialogs, would it make sense to also show the warning if the deleted thing is a customer managed environment? We also do that in the edit/deploy components in that case.

frontend/ui/src/app/deployments/deployment-targets.component.html
```diff
@@ -181,21 +181,68 @@
...
-                  aria-label="Delete"
-                  (click)="deleteDeploymentTarget(dt)"
-                  class="py-2 px-3 text-red-700 hover:text-white border border-red-700 hover:bg-red-800 focus:ring-4 focus:outline-none focus:ring-red-300 font-medium rounded-lg text-sm text-center dark:border-red-500 dark:text-red-500 dark:hover:text-white dark:hover:bg-red-600 dark:focus:ring-red-900">
-                  <fa-icon [icon]="faTrash" class="h-4 w-4"></fa-icon>
+                  aria-label="More actions"
+                  class="p-2 flex items-center text-sm font-medium text-center text-gray-900 focus:outline-none bg-white rounded-lg border border-gray-200 hover:bg-gray-100 hover:text-primary-700 focus:z-10 focus:ring-4 focus:ring-gray-200 dark:focus:ring-gray-700 dark:bg-gray-800 dark:text-gray-400 dark:border-gray-600 dark:hover:text-white dark:hover:bg-gray-700"
+                  cdkOverlayOrigin
+                  #dropdownTrigger="cdkOverlayOrigin"
+                  (click)="showDropdownForId = dt.id">
+                  <fa-icon [icon]="faEllipsisVertical" class="h-5 w-5 text-gray-500 dark:text-gray-400"></fa-icon>
```
christophenne, line 192: this dropdown is a good idea :+1: 

I only found one weird bug (very minor): When closing the dropdown with the escape key, you can't reopen the same dropdown anymore. You have to open another one first and then it is openable again. :melting_face:

christophenne: Just a wild guess, but maybe the `showDropdownForId` isn't cleaned in that case? In the license edit version dropdown, I implemented it with a signal and an effect that runs if visibility changes, such that you can react on every kind of close: https://github.com/glasskube/distr/blob/main/frontend/ui/src/app/licenses/edit-license.component.ts#L112

The use case is different there though, and the escape closes the whole drawer in that case :see_no_evil:

frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -213,6 +230,30 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
       .subscribe();
   }
 
+  async deleteDeployment(dt: DeploymentTarget, d: Deployment) {
+    if (d.id) {
+      if (
+        await firstValueFrom(
+          this.overlay.confirm(`
```
christophenne: We should maybe invest in a "generic confirm modal" in the future, where we can put a whole template instead of just text. This is a bit too wall-of-texty in my opinion.

frontend/ui/src/app/deployments/deployment-targets.component.html
```diff
@@ -181,21 +181,68 @@
...
+                            aria-label="Delete"
+                            (click)="showDropdownForId = undefined; deleteDeployment(dt, dt.deployment)"
+                            class="text-start block w-full px-4 py-2 text-sm text-red-700 hover:text-white hover:bg-red-800 dark:text-red-500 dark:hover:text-white dark:hover:bg-red-600">
+                            <fa-icon [icon]="faTriangleExclamation" class="inline-block w-4 mr-2"></fa-icon>
+                            Undeploy
+                          </button>
+                        </li>
+                      }
+                      <li>
+                        <button
```
christophenne, line 235: Would it make sense to make target deletion only possible if there is no deployment inside?

internal/handlers/deployments.go
```diff
@@ -66,6 +67,38 @@ func putDeployment(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func deleteDeploymentHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
```
christophenne, line 71: Is there any benefit in returning a `http.HandlerFunc` instead of just implementing the `http.Handler`? Not that I think it should change, just out of curiosity.

internal/handlers/deployments.go
```diff
@@ -66,6 +67,38 @@ func putDeployment(w http.ResponseWriter, r *http.Request) {
...
+			target, err := db.GetDeploymentTargetForDeploymentID(ctx, deployment.ID)
+			if err != nil {
+				log.Warn("could not get DeploymentTarget", zap.Error(err))
+				sentry.GetHubFromContext(ctx).CaptureException(err)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return err
+			}
+			if target.OrganizationID != orgId ||
+				(*auth.CurrentUserRole() != types.UserRoleVendor && target.CreatedByUserAccountID != auth.CurrentUserID()) {
+				http.NotFound(w, r)
```
christophenne, line 87: should return here I think

cmd/agent/docker/agent_deployment.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		return nil, err
+	} else {
+		result := make([]AgentDeployment, 0, len(entries))
+		for _, entry := range entries {
+			if !entry.IsDir() {
+				if file, err := os.Open(path.Join(agentDeploymentDir(), entry.Name())); err != nil {
+					return nil, err
+				} else {
+					defer file.Close()
```
christophenne: `defer` inside the for loop will make the open files pile up, maybe move the open+parse+close part to a function. Realistically won't make a huge difference of course.

cmd/agent/docker/config.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import "os"
+
+func ScratchDir() string {
+	if dir := os.Getenv("DISTR_AGENT_SCRATCH_DIR"); dir != "" {
+		return dir
+	}
+	return "./scratch"
```
christophenne, line 9: bit confused, should it not be "/scratch"?

frontend/ui/src/app/deployments/deployment-targets.component.ts
kosmoz: I agree, something inbetween the overlay and confirm modal is definitely needed, but I thought this would make sense as a separate change set

cmd/agent/docker/docker_actions.go
kosmoz: For helm, this is not generally possible, unless a PVC was created by a controller that isn't configured to clean it up (e.g. StatefulSet). For docker we could simply remove the `--volumes` flag but then the feature would work differently on both platforms.

kosmoz: Not really sure what's best, but maybe @pmig can share their opinion?

kosmoz: More difficult deletion confirmation makes sense in any case though. But maybe also as a separate change set?

frontend/ui/src/app/deployments/deployment-targets.component.html
kosmoz: @pmig ?

internal/handlers/deployments.go
kosmoz: There are two benefits, both very small and kind of subjective:
1. It's easier to change if the need arises to somehow parametrize the handler (e.g. [here](https://github.com/glasskube/distr/blob/9a11de50e287b21fa3ff98a87d01596b24b640ba/internal/handlers/applications.go#L273))
2. It works better with my autocomplete :^)

cmd/agent/docker/config.go
kosmoz: this is for local dev mostly, if running the agent outside of a container it probably can't write to /scratch. we always set the env var in the docker compose file so it doesn't impact a deployed agent

frontend/ui/src/app/deployments/deployment-targets.component.html
kosmoz: I solved this by adding `(detach)="showDropdownForId = undefined"` to the `cdkConnectedOverlay`.

frontend/ui/src/app/deployments/deployment-targets.component.html
pmig: yes 100% with the un deploy feature it should be possible :+1:

cmd/agent/docker/docker_actions.go
pmig: I think making more difficult to delete is a good thing, maybe we can hint that a user should export / backup the docker containers with `docker save`?

internal/handlers/deployments.go
christophenne: Ok sounds good

frontend/ui/src/app/deployments/deployment-targets.component.ts
christophenne: sounds good :+1:

frontend/ui/src/app/deployments/deployment-targets.component.html
christophenne: ok even better

cmd/agent/docker/docker_actions.go
kosmoz: https://linear.app/glasskube/issue/GK-553/make-it-more-difficult-to-delete-things-when-important-data-might-be

frontend/ui/src/app/deployments/deployment-targets.component.html
```diff
@@ -181,21 +181,71 @@
...
+                          type="button"
+                          aria-label="Delete"
+                          [disabled]="dt.deployment"
+                          [title]="dt.deployment ? 'Undeploy application first' : undefined"
```
christophenne:
```suggestion
                          [title]="dt.deployment ? 'Undeploy application first' : ''"
```

shows "undefined" otherwise :sweat_smile:

PR #523, opened 2025-02-19
fix: fix error on organization branding creation

pmig:
ID is only used by the frontend to determine if the update or creation endpoint should be called

### DIFF
--- a/internal/types/organization.go
+++ b/internal/types/organization.go
@@ -31,7 +31,7 @@ type OrganizationWithBranding struct {
 }
 
 type OrganizationBranding struct {
-	ID                     *uuid.UUID `db:"id" json:"id"`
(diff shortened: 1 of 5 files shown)


internal/types/organization.go
```diff
@@ -31,7 +31,7 @@ type OrganizationWithBranding struct {
 }
 
 type OrganizationBranding struct {
-	ID                     uuid.UUID  `db:"id" json:"id"`
+	ID                     *uuid.UUID `db:"id" json:"id"`
```
kosmoz: Is this change intentional? Because we never user pointers for the ID in any other DB models…

pmig: No, recerted

kosmoz (approved):

PR #525, opened 2025-02-19
chore: send http exceptions to Sentry

pmig:


frontend/ui/src/app/services/error-toast.interceptor.ts
```diff
@@ -13,6 +15,10 @@ export const errorToastInterceptor: HttpInterceptorFn = (req, next) => {
         if (msg) {
           toast.error(msg);
         }
+
+        if (err.status && (err.status === 400 || err.status >= 500)) {
```
christophenne, line 19: Internal server errors (>= 500) are tracked via backend already, so there would be lots of duplicates I think?

pmig: Yes on purpose. We might get additional context. I hope Sentry somehow aggregates them.

PR #535, opened 2025-02-20
fix(ui): make autotrim work with FormControls

kosmoz:


frontend/ui/src/app/directives/autotrim.directive.ts
```diff
@@ -1,11 +1,17 @@
...
 @Directive({selector: 'input[autotrim]'})
 export class AutotrimDirective {
+  private readonly ngControl = inject(NgControl, {optional: true});
+
   @HostListener('blur', ['$event']) onBlur(event: Event) {
     const target = event.target as HTMLInputElement;
     if (target.value !== target.value.trim()) {
       target.value = target.value.trim();
     }
+    if (typeof this.ngControl?.value === 'string' && this.ngControl.value != this.ngControl.value.trim()) {
```
christophenne:
```suggestion
    if (typeof this.ngControl?.value === 'string' && this.ngControl.value !== this.ngControl.value.trim()) {
```

PR #550, opened 2025-02-24
feat(frontend): add indicator if update is available

kosmoz:


frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -412,5 +448,39 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
...
+
+  private findMaxVersion(versions: ApplicationVersion[]): ApplicationVersion | undefined {
+    try {
+      return maxBy(
+        versions,
+        (version) => new SemVer(version.name),
+        (a, b) => a.compare(b) > 0
+      );
+    } catch (e) {
+      if (!environment.production) {
```
christophenne: I think the warning would also be fine to have in production. 

Also, we should probably document that logic somewhere.

frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -19,13 +19,16 @@ import {
...
+  ApplicationVersion,
   Deployment,
   DeploymentRequest,
   DeploymentRevisionStatus,
   DeploymentTarget,
   DeploymentTargetScope,
   DeploymentType,
+  DeploymentWithLatestRevision,
 } from '@glasskube/distr-sdk';
+import dayjs from 'dayjs';
```
christophenne:
```suggestion
```

unused import

frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -161,6 +173,26 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
 
   statuses: Observable<DeploymentRevisionStatus[]> = EMPTY;
 
+  protected deploymentTargetsWithUpdate$: Observable<{dt: DeploymentTarget; version: ApplicationVersion}[]> =
+    this.deploymentTargets$.pipe(
```
christophenne, line 175: No strong opinion, but it could also only show updates of the filtered (== currently shown targets). But I wouldn't expect it from other applications, so this should be fine too. (Also makes sense like this, because the update banner is above or outside of the filtered context (visually).

frontend/ui/src/app/deployments/deployment-targets.component.ts
kosmoz: fair, changed it

you mean in the docs? or as part of the app somewhere?

frontend/ui/src/app/deployments/deployment-targets.component.ts
kosmoz: I was contemplating this too, but my gut told me that this would be more intuitive. 

Would keep as is for now, maybe we want to change it later

frontend/ui/src/app/deployments/deployment-targets.component.ts
christophenne: I think somewhere in the docs should be fine.

PR #556, opened 2025-02-25
feat(ui): add beta feature activation tooltip

pmig:


kosmoz (commented):
The overlay appears on hover but only disappears on click. Is this intended?

pmig:
> The overlay appears on hover but only disappears on click. Is this intended?

Yes, so users can click the link and the overlay is more prominent

kosmoz:
> > The overlay appears on hover but only disappears on click. Is this intended?
> 
> Yes, so users can click the link and the overlay is more prominent

Can we not show it on click? I think it's really annoying the way it is

kosmoz (dismissed):
> > > The overlay appears on hover but only disappears on click. Is this intended?
> > 
> > Yes, so users can click the link and the overlay is more prominent
> 
> Can we not show it on click? I think it's really annoying the way it is

Your call at the end of the day, but I would strongly suggest changing it

kosmoz (approved):

PR #574, opened 2025-02-27
chore(ui): make cache resettable

christophenne:


frontend/ui/src/app/services/cache.ts
```diff
@@ -1,46 +1,71 @@
...
 import {compareBy, distinctBy, Predicate} from '../../util/arrays';
 import {BaseModel, Named} from '@glasskube/distr-sdk';
 
+interface ReactiveListEvent<T> {
+  objects?: T[]; // used for type 'reset'
+  object?: T; // used when type is 'save' or 'remove'
+  type: ReactiveListEventType;
+}
+
+type ReactiveListEventType = 'save' | 'remove' | 'reset';
```
kosmoz: I think this could be improved slightly using a union type, something like this (I didn't try it):
```ts
type ReactiveListEvent<T> = 
  | {type: 'save' | 'remove', object: T}
  | {type: 'reset', objects: T[]};
```

This way, we could rely on the type system to ensure that object/objects is set appropriately and not undefined where it shouldn't be, simplifying the if expression below.

EDIT: Apart from that, looking good! I don't think the potential race condition will be an issue any time soon, we don't have to handle it now but it's good to keep in mind so let's keep the comment I would say

christophenne: agree, thanks

kosmoz (approved):

PR #578, opened 2025-02-28
chore(backend): application of existing deployment should not be changeable

christophenne:


internal/handlers/deployments.go
```diff
@@ -163,6 +163,9 @@ func validateDeploymentRequest(
 		} else if *request.ApplicationLicenseID != *target.Deployment.ApplicationLicenseID {
 			return badRequestError(w, "can not update license")
 		}
+		if target.Deployment.ApplicationID != (*app).ID {
```
kosmoz: I think this deref is not necessary
```suggestion
		if target.Deployment.ApplicationID != app.ID {
```

kosmoz (changes requested):

kosmoz (approved):

PR #583, opened 2025-02-28
chore(backend): add `SERVER_SHUTDOWN_DELAY_DURATION` config variable

kosmoz:
it does exactly what it sounds it does.

This is needed for our AWS deployment where a way to delay the container termination is required. because we use a distroless base image, the usual pattern of calling `sleep` in a preStop lifecycle hook does not work in this case.

### DIFF
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -128,7 +128,7 @@ func init() {
 	}
 
 	if value, ok := os.LookupEnv("SERVER_SHUTDOWN_DELAY_DURATION"); ok {
-		serverShutdownDelayDuration = util.PtrTo(util.Require(time.ParseDuration(value)))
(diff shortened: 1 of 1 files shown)


internal/env/env.go
```diff
@@ -125,6 +126,10 @@ func init() {
 	if value, ok := os.LookupEnv("USER_EMAIL_VERIFICATION_REQUIRED"); ok {
 		userEmailVerificationRequired = util.Require(strconv.ParseBool(value))
 	}
+
+	if value, ok := os.LookupEnv("SERVER_SHUTDOWN_DELAY_DURATION"); ok {
+		serverShutdownDelayDuration = util.PtrTo(util.Require(time.ParseDuration(value)))
```
christophenne: just saw there is also a `requirePositiveDuration` helper function already

PR #598, opened 2025-03-03
chore(ui): require additional input when deleting deployments

christophenne:


frontend/ui/src/app/services/overlay.service.ts
```diff
@@ -54,7 +58,7 @@ export class OverlayService {
   private readonly viewportRuler = inject(ViewportRuler);
   private readonly viewContainerRef = inject(ViewContainerRef);
 
-  public confirm(messageOrConfig: string | ConfirmMessage) {
+  public confirm(messageOrConfig: ConfirmConfig | string) {
     const config = typeof messageOrConfig === 'string' ? {message: messageOrConfig} : messageOrConfig;
```
kosmoz:
```suggestion
  public confirm(messageOrConfig: ConfirmConfig | string) {
    const config: ConfirmConfig =
      typeof messageOrConfig === 'string' ? {message: {message: messageOrConfig}} : messageOrConfig;
```

kosmoz (changes requested):
Found a small bug (see suggestion).

I think it's quite good, but I have two suggestions:
1. Would it make sense to use different messages? At least different messages for different actions like DELETE and UNDEPLOY or something like that
2. I kind of expected to be able to "submit" the deletion by pressing ENTER, but I'm not 100% sure that would be desireable (@pmig?)

pmig:
> 2. I kind of expected to be able to "submit" the deletion by pressing ENTER, but I'm not 100% sure that would be desirable (@pmig?)

If the `DELETE` was typed in I think form submission with the enter button is fine - no must though.

kosmoz:
> I tried the Enter approach on other confirm modals (like deleting application) and it opens a modal on top of the modal, probably because the opening button still has the focus 🫠

That's wild, but I guess it's not that important anyways

PR #601, opened 2025-03-05
feat(ui): multiple artifacts

christophenne:


frontend/ui/src/app/artifacts/artifact-licenses/artifact-licenses.component.ts
```diff
@@ -140,4 +141,10 @@ export class ArtifactLicensesComponent implements OnDestroy {
   protected readonly faPen = faPen;
   protected readonly faTrash = faTrash;
   protected readonly faXmark = faXmark;
+
+  ngOnInit(): void {
```
kosmoz: probably debug remnant

kosmoz (commented):
hate to be that guy but… the two selects have different looking chevrons:
image

also, for now I think it's fine but I think it would make sense to (1) only allow to add artifacts not already in the list and (2) disable changing an artifact already in the list (apart from adding/removing tags and removing it)

kosmoz (approved):

PR #602, opened 2025-03-05
feat: add Docker Swarm support as a new deployment package

asherAbecasiss:
This PR introduces a new **Docker Swarm deployment** package alongside the existing Docker Compose deployment. 
The Swarm package enables running services in a Swarm cluster while keeping the current Compose-based workflow intact.

### DIFF
--- a/cmd/agent/docker-swarm/docker_actions.go
+++ b/cmd/agent/docker-swarm/docker_actions.go
@@ -44,14 +44,6 @@ func parseEnvFile(envData []byte) (map[string]string, error) {
 	}
 	return envVars, scanner.Err()
 }
-func replaceEnvVars(composeData []byte, envVars map[string]string) []byte {
-	content := string(composeData)
-	for key, value := range envVars {
-		placeholder := fmt.Sprintf("${%s}", key)
-		content = strings.ReplaceAll(content, placeholder, value)
-	}
-	return []byte(content)
-}
 
 func ApplyComposeFileSwarm(
 	ctx context.Context,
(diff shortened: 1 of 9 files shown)


cmd/agent/docker-swarm/docker_actions.go
```diff
@@ -0,0 +1,121 @@
...
+			return nil, fmt.Errorf("invalid environment variable: %s", line)
+		}
+		envVars[parts[0]] = parts[1]
+	}
+	return envVars, scanner.Err()
+}
+func replaceEnvVars(composeData []byte, envVars map[string]string) []byte {
+	content := string(composeData)
+	for key, value := range envVars {
+		placeholder := fmt.Sprintf("${%s}", key)
```
kosmoz: Parsing the env file and replacing variables like this seems kind of brittle. I think understand that `docker stack` doesn't support variable substitution directly, but a better way appears to be using `docker compose config` to perform variable substitution and other generators and run `docker stack deploy` with the output of that ([source](https://github.com/moby/moby/issues/29133#issuecomment-442912378)). Do you think that would work?

kosmoz (commented):
Hi Asher, thanks for contributing to Distr!! :rocket: 

I think the idea of supporting docker swarm deployments is great, however I have some concerns regarding the implementation at hand:

At a glance, it appears that there is a lot of overlap with the already existing docker compose agent. Only the code in `docker_actions.go` appears genuinely new. For this reason and because `docker stack deploy` seems like just another "mode" of docker deployment similar to docker compose, I think it would be preferable to have one agent support both and toggle between them using a new property in the agent resource response. A user could then decide between compose and swarm when creating a deployment target. Currently, I don't think there is a way for a user to use this feature so some kind support in the UI/backend would be needed in any case.

Also, we'd be happy to learn more about your use-case in order to fully understand the requirements here. :slightly_smiling_face:

cmd/agent/docker-swarm/docker_actions.go
asherAbecasiss: Sure, let me look into this.

asherAbecasiss: I previously used this script, and it was incredibly helpful for managing env files with Docker Swarm. I just implemented it.

```
start_with_env_file() {
    
    stack=${1:-${PWD##*/}} # by default, the name of the cointaining folder
    compose_file=${2:-docker-compose.yml}

    if [ ! -f $compose_file ]; then
        echo "Misses compose file: $compose_file" >&2
        return 1
    fi

    # execute as a subcommand in order to avoid the variables remain set
    (
        # export variables excluding comments
        [ -f .env ] && export $(sed '/^#/d' .env)

        # Use dsd your_stack your_compose_file to override the defaults
        docker stack deploy --compose-file $compose_file "home"
    )
}
```

pmig:
Hi @asherAbecasiss thanks again for creating your PR.

I quickly wanted to check in if you got docker swarm support working locally by building and re-tagging our docker-swarm agent.

In order for us to implement docker swarm (stack deploy) support into Distr, we would make a bigger refactoring of the pr and implementing the changes into the existing docker agent and adding the necessary changes in the UI and database to support docker swarm deployments nativley.

Let us know how your tests have been and if it make sense for us to build docker swarm support into upstream or you are happy with maintaining your fork with your custom agent.

I also created a discusion (https://github.com/glasskube/distr/discussions/609) and shared it on Discord to see if anyone else is interested in Docker swarm support.

asherAbecasiss:
Hi @pmig ,

Thanks for reviewing the PR and for the follow-up!

Yes, I was able to get Docker Swarm support working locally by building and re-tagging the Docker Swarm agent. The deployment process worked as expected, and I validated the stack deployment behavior in a Swarm environment.

I understand that integrating Docker Swarm support into Distr would require a broader refactoring to align with the existing Docker agent, UI, and database structure. If there is interest in upstream support, I’d be happy to collaborate and adjust the implementation accordingly.

pmig:
> I understand that integrating Docker Swarm support into Distr would require a broader refactoring to align with the existing Docker agent, UI, and database structure. If there is interest in upstream support, I’d be happy to collaborate and adjust the implementation accordingly.

**The needed steps would be:**

- [ ] Relabel the "Docker Compose" to "Docker" deployment method (also for file uploads).
- [ ] Introduce a "DockerType" similar to the "HelmChartType" that can be either "compose" or "swarm" (or "stack"?). "Compose" should be the default, and a migration should be performed.
- [ ] Use the same Docker agent in both cases and just adapt the execution based on that type (similar to how the Kubernetes agent does it).
- [ ] Adapt the environment parsing based on @kosmoz  [suggestion](https://github.com/glasskube/distr/pull/602#discussion_r1981524321).

pmig:
closed in favor of: https://github.com/glasskube/distr/pull/645

PR #608, opened 2025-03-06
chore: support registration mode via environment

christophenne:


internal/handlers/auth.go
```diff
@@ -69,9 +74,24 @@ func authLoginHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func authRegisterGetHandler() http.HandlerFunc {
+	ok := env.Registration() == env.RegistrationEnabled
```
kosmoz, line 78: would it make sense to add this to the frontend environment handler instead?

internal/env/env.go
```diff
@@ -33,6 +33,7 @@ var (
 	frontendPosthogUIHost         *string
 	userEmailVerificationRequired = true
 	serverShutdownDelayDuration   *time.Duration
+	registration                  = RegistrationHidden
```
kosmoz: I think it would make sense to use Enabled as the default. If I remember correctly, one of the reason this feature is needed is because users trying to selfhost don't know how to create an account.

internal/handlers/auth.go
christophenne: I had it like that, but I saw two "problems"/gotchas:
* the `internal/environment` response is cached in `sessionStorage`, so to observe a change of the variable you have to clear it or open a new tab. 
* when it is not cached, the endpoint is mostly called twice because the first response isn't cached yet, and I didn't really want to touch this initial out-of-Angular part too much. 

Both are not big blockers and I can also change it back. I agree that this is more of a part of that environment call.

internal/env/env.go
christophenne: True, will change it.

internal/handlers/auth.go
kosmoz: ok fine for now

PR #611, opened 2025-03-07
fix: only return licensed application if feature is active

pmig:


internal/db/applications.go
```diff
@@ -23,6 +23,17 @@ const (
 			FROM applicationversion av
 			WHERE av.application_id = a.id
 		), array[]::record[]) as versions `
+
+	applicationWithLicensedVersionsOutputExpr = applicationOutputExpr + `,
```
kosmoz, line 29: Maybe I'm missing something, but I think this doesn't check the application license owner

internal/db/applications.go
```diff
@@ -90,12 +101,11 @@ func GetApplicationsByOrgID(ctx context.Context, orgID uuid.UUID) ([]types.Appli
...
 func GetApplicationsWithLicenseOwnerID(ctx context.Context, id uuid.UUID) ([]types.Application, error) {
 	db := internalctx.GetDb(ctx)
-	// TODO: Only include versions from at least one license
 	if rows, err := db.Query(ctx, `
-			SELECT DISTINCT `+applicationWithVersionsOutputExpr+`
+			SELECT DISTINCT `+applicationWithLicensedVersionsOutputExpr+`
 			FROM ApplicationLicense al
 				LEFT JOIN Application a ON al.application_id = a.id
-			WHERE al.owner_useraccount_id = @id
+			WHERE al.owner_useraccount_id = @id AND (al.expires_at IS NULL OR al.expires_at > now())
```
kosmoz, line 140: I would 
1. do this in a separate PR
2. probably do it in application code and show it as "expired" or something instead, hiding like this could cause unforeseen problems

internal/db/applications.go
kosmoz, line 29: Also please use all caps for SQL keywords always

internal/db/applications.go
pmig: this only applies if the license feature is active and the user is not a `vendor` (see application handler)

pmig: (and the license is already checked, only the expiration date was ignored)

internal/db/applications.go
pmig: only `ApplicationLicense`s `al`  are in the context if the current user is the owner, see https://github.com/glasskube/distr/blob/main/internal/db/applications.go#L98

kosmoz: thanks for clarifying

internal/db/applications.go
kosmoz: OK, but we let's be mindful of this in the future

internal/db/applications.go
```diff
@@ -126,6 +136,24 @@ func GetApplication(ctx context.Context, id, orgID uuid.UUID) (*types.Applicatio
...
+			SELECT DISTINCT `+applicationWithLicensedVersionsOutputExpr+`
+			FROM ApplicationLicense al
+				LEFT JOIN Application a ON al.application_id = a.id
+			WHERE al.owner_useraccount_id = @ownerID AND a.id = @id AND (al.expires_at IS NULL OR al.expires_at > now())
+			ORDER BY a.name
+			`, pgx.NamedArgs{"ownerID": oID, "id": id}); err != nil {
+		return nil, fmt.Errorf("failed to query applications: %w", err)
+	} else if application, err :=
+		pgx.CollectExactlyOneRow(rows, pgx.RowToStructByName[types.Application]); err != nil {
+		return nil, fmt.Errorf("failed to get application: %w", err)
```
kosmoz, line 186: To return the correct status code when the application doesn't exist, also handle the error in the handler
```suggestion
		if errors.Is(err, pgx.ErrNoRows) {
			return nil, apierrors.ErrNotFound
		}
		return nil, fmt.Errorf("failed to get application: %w", err)
```

internal/handlers/applications.go
```diff
@@ -140,7 +140,38 @@ func getApplications(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if org.HasFeature(types.FeatureLicensing) && *auth.CurrentUserRole() == types.UserRoleCustomer {
+
+		if applicationID, err := uuid.Parse(r.PathValue("applicationId")); err != nil {
+			http.NotFound(w, r)
+			return
+		} else {
+			application, err := db.GetApplicationWithLicenseOwnerID(ctx, auth.CurrentUserID(), applicationID)
+			if err != nil {
```
kosmoz:
```suggestion
			if errors.Is(err, apierrors.NotFound) {
				http.NotFound(w, r)
			} else if err != nil {
```

internal/db/applications.go
```diff
@@ -90,12 +101,11 @@ func GetApplicationsByOrgID(ctx context.Context, orgID uuid.UUID) ([]types.Appli
 
 func GetApplicationsWithLicenseOwnerID(ctx context.Context, id uuid.UUID) ([]types.Application, error) {
 	db := internalctx.GetDb(ctx)
-	// TODO: Only include versions from at least one license
 	if rows, err := db.Query(ctx, `
-			SELECT DISTINCT `+applicationWithVersionsOutputExpr+`
+			SELECT DISTINCT `+applicationWithLicensedVersionsOutputExpr+`
```
kosmoz, line 137: there is a subtle bug here: if a customer has two licenses for the same application that have different restrictions they get that app multiple times instead of the versions list being the aggregate of licensed versions

pmig: great catch added a workaround, this can be refactored at a later point

PR #622, opened 2025-03-12
feat: add db manifest handler

kosmoz:


internal/registry/manifest.go
```diff
@@ -553,6 +314,196 @@ func (m *manifests) handleReferrers(resp http.ResponseWriter, req *http.Request)
...
+	resp.WriteHeader(http.StatusOK)
+	return nil
+}
+
+func (handler *manifests) handlePut(resp http.ResponseWriter, req *http.Request, repo, target string) *regError {
+	buf := &bytes.Buffer{}
+	if _, err := io.Copy(buf, req.Body); err != nil {
+		return regErrInternal(err)
+	}
+	manifestDigest, _, _ := v1.SHA256(bytes.NewReader(buf.Bytes()))
```
christophenne, line 399: could use some error handling here I think

internal/registry/manifest.go
```diff
@@ -425,45 +205,42 @@ func (m *manifests) handleCatalog(resp http.ResponseWriter, req *http.Request) *
```
christophenne, line 205: probably TODO: only return accessible repos

kosmoz: This should already be done by the manifest handler, see 

https://github.com/glasskube/distr/blob/a6844b2f5a4c102113c84f84815be03a30a0ceae/internal/registry/manifest/db/handler.go#L48-L71

internal/registry/manifest.go
kosmoz: true, although to be fair this is part of the original code :sweat_smile:

PR #623, opened 2025-03-12
feat: add organization slug and organization settings

pmig:


pmig:
superceeded by https://github.com/glasskube/distr/pull/625

PR #625, opened 2025-03-12
feat: add organization slug and organization settings

pmig:


frontend/ui/src/app/app.routes.ts
```diff
@@ -197,6 +198,12 @@ export const routes: Routes = [
             data: {userRole: 'vendor'},
             canActivate: [requiredRoleGuard('vendor')],
           },
+          {
```
christophenne, line 201: optional, but it might make sense to prefix with `organization-` or even do a common base route `organization/` for settings and branding.

frontend/ui/src/app/organization-settings/organization-settings.component.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+      this.form.patchValue({
+        name: this.organization.name,
+        slug: this.organization.slug,
+      });
+    } catch (e) {
+      const msg = getFormDisplayedError(e);
+      if (msg && e instanceof HttpErrorResponse && e.status !== 404) {
+        // it's a valid use case for an organization to have no branding (hence 404 is not shown in toast)
+        this.toast.error(msg);
```
christophenne:
```suggestion
      if (msg) {
        this.toast.error(msg);
```

internal/db/organization.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/jackc/pgx/v5"
 )
 
+const (
+	organizationOutputExpr = `
+		o.id, o.created_at, o.name, o.slug, o.features
+`
```
christophenne:
```suggestion
	organizationOutputExpr = ` o.id, o.created_at, o.name, o.slug, o.features `
```

internal/db/organization.go
```diff
@@ -30,10 +36,28 @@ func CreateOrganization(ctx context.Context, org *types.Organization) error {
 	}
 }
 
+func UpdateOrganization(ctx context.Context, org *types.Organization) error {
+	db := internalctx.GetDb(ctx)
+	rows, err := db.Query(ctx,
+		"UPDATE Organization AS o SET name = @name, slug = @slug WHERE id = @id RETURNING "+organizationOutputExpr,
+		pgx.NamedArgs{"id": org.ID, "name": org.Name, "slug": org.Slug},
+	)
+	if err != nil {
```
christophenne, line 46: optional, but you could check whether the error is a unique contraint violation, and pass that back to the UI, so that you can show "Slug already taken" or something like that.

kosmoz: yes, please do this! we already have some examples for exactly this, e.g.:

https://github.com/glasskube/distr/blob/a6844b2f5a4c102113c84f84815be03a30a0ceae/internal/db/applications.go#L267-L269

pmig:
> L(mostly)GTM!
> 
> But: Other than the minor comments, user registration is not working anymore: ` handlers/auth.go:123 user registration failed {"requestId": "christoph-gk/QEDn1JQ10J-000096", "error": "cannot find field slug in returned row"}`
> 
> `db.CreateOrganization` also needs to return the slug.
> 
> Also, for a newly created organization, I can't update the settings without setting the slug, with some very weird behaviiour (the request is not sent and a toast "Something went wrong" shows up) – haven't debugged it.

Thanks for testing, also fixed the signup and updating settings without the slug (an undefined slug couldn't get trimmed)

frontend/ui/src/app/app.routes.ts
pmig: I understand the concern, but wouldn't change it as of now.

internal/handlers/organization.go
```diff
@@ -35,3 +41,54 @@ func getOrganization(w http.ResponseWriter, r *http.Request) {
...
+		if organization.ID == uuid.Nil {
+			organization.ID = existingOrganization.ID
+		} else if organization.ID != existingOrganization.ID {
+			http.Error(w, "organization id does not match", http.StatusBadRequest)
+			return
+		}
+
+		if err := db.UpdateOrganization(ctx, &organization); errors.Is(err, apierrors.ErrConflict) {
+			http.Error(w, "Slug is not available", http.StatusBadRequest)
+			return
```
kosmoz:
```suggestion
```

internal/handlers/organization.go
```diff
@@ -35,3 +41,54 @@ func getOrganization(w http.ResponseWriter, r *http.Request) {
...
+
+	organization, err := JsonBody[types.Organization](w, r)
+	if err != nil {
+		return
+	} else if organization.Name == "" {
+		http.Error(w, "name is required", http.StatusBadRequest)
+		return
+	}
+
+	if existingOrganization, err := db.GetOrganizationByID(ctx, *auth.CurrentOrgID()); err == nil {
```
kosmoz: It's more conventional to check errors the other way around, i.e. `err != nil`

internal/handlers/organization.go
```diff
@@ -35,3 +41,54 @@ func getOrganization(w http.ResponseWriter, r *http.Request) {
...
+		if existingOrganization.Slug != nil {
+			if organization.Slug == nil || *organization.Slug == "" {
+				http.Error(w, "Slug can not get unset", http.StatusBadRequest)
+				return
+			}
+
+			if matched, _ := regexp.MatchString("^[a-z]+$", *organization.Slug); !matched {
```
kosmoz: The distribution spec has this pattern for the manifest `<name>`: 

`[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*`

We could use the first part (`[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*`) for the organization slug.

Also, I think it would make sense to limit the length. 64 characters? 256 characters?

internal/handlers/organization.go
```diff
@@ -54,18 +54,25 @@ func updateOrganization(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusInternalServerError)
+	} else {
 		if existingOrganization.Slug != nil {
 			if organization.Slug == nil || *organization.Slug == "" {
 				http.Error(w, "Slug can not get unset", http.StatusBadRequest)
 				return
 			}
-
-			if matched, _ := regexp.MatchString("^[a-z]+$", *organization.Slug); !matched {
+			slugPattern := "^[a-z0-9]+((\\.|_|__|-+)[a-z0-9]+)*$"
```
christophenne: just noticed this now, but the pattern and length validation need to be done outside of the `existingOrganization.Slug` - `if`, I think

frontend/ui/src/app/organization-settings/organization-settings.component.html
```diff
@@ -0,0 +1,75 @@
...
+                    autotrim
+                    type="text"
+                    name="slug"
+                    id="slug"
+                    class="bg-gray-50 border border-gray-300 text-sm text-gray-900 rounded-lg focus:ring-primary-600 focus:border-primary-600 block w-full p-2.5 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500"
+                    placeholder="slug"
+                    [pattern]="slugPattern"
+                    [maxlength]="slugMaxLength" />
+                  <p class="mt-1 mb-3 text-xs font-normal text-gray-500 dark:text-gray-400">
+                    The slug will used as part of the url for artifact registry. It must be only lower cased letters.
```
kosmoz:
```suggestion
                    The slug will used as part of the url for artifact registry. It must start with a lower case letter
                    or digit and consist only of lowercase letters, digits, as well as the following special characters:
                    “.”, “_”, “__”, “-”.
```

frontend/ui/src/app/organization-settings/organization-settings.component.html
```diff
@@ -0,0 +1,75 @@
...
+                  <input
+                    formControlName="slug"
+                    autotrim
+                    type="text"
+                    name="slug"
+                    id="slug"
+                    class="bg-gray-50 border border-gray-300 text-sm text-gray-900 rounded-lg focus:ring-primary-600 focus:border-primary-600 block w-full p-2.5 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500"
+                    placeholder="slug"
+                    [pattern]="slugPattern"
+                    [maxlength]="slugMaxLength" />
```
kosmoz: These don't do anything when you use reactive form, see https://angular.dev/guide/forms/form-validation#interaction-with-native-html-form-validation
```suggestion
                    placeholder="slug" />
```

kosmoz (approved):

PR #624, opened 2025-03-12
feat: show registry artifacts in UI

christophenne:

PR #632, opened 2025-03-13
feat: add checking license for artifact server

kosmoz:


internal/db/artifacts.go
```diff
@@ -165,6 +165,106 @@ func GetLicensedArtifactVersions(
...
+				FROM ArtifactVersion av
+				JOIN ArtifactVersionPart avp ON av.id = avp.artifact_version_id
+				JOIN ArtifactVersionAggregate agg ON avp.artifact_blob_digest = agg.manifest_blob_digest
+		)
+		SELECT exists(
+			SELECT *
+				FROM ArtifactVersionAggregate av
+				JOIN ArtifactLicense_Artifact ala ON ala.artifact_version_id = av.id
+				JOIN ArtifactLicense al ON ala.artifact_license_id = al.id
+				WHERE al.owner_useraccount_id = @userId
```
christophenne, line 281: at all points where the owner is checked, there also needs to be ` AND (al.expires_at IS NULL OR al.expires_at > now())`, but expires_at only is introduced in the #624 PR, so it can be done afterwards

internal/db/artifacts.go
```diff
@@ -165,6 +165,106 @@ func GetLicensedArtifactVersions(
...
+			UNION ALL
+			SELECT DISTINCT av.id, av.manifest_blob_digest
+				FROM ArtifactVersion av
+				JOIN ArtifactVersionPart avp ON av.id = avp.artifact_version_id
+				JOIN ArtifactVersionAggregate agg ON avp.artifact_blob_digest = agg.manifest_blob_digest
+		)
+		SELECT exists(
+			SELECT *
+				FROM ArtifactVersionAggregate av
+				JOIN ArtifactLicense_Artifact ala ON ala.artifact_version_id = av.id
```
christophenne: a license could also have a `NULL` version ID to license all versions of the artifact, I think this is not yet represented in this check

kosmoz: fair

PR #639, opened 2025-03-14
chore: use org slug for artifact names

kosmoz:


internal/registry/authz/authorizer.go
```diff
@@ -43,7 +43,7 @@ func (a *authorizer) Authorize(ctx context.Context, nameStr string, action Actio
 		return errors.New("name invalid")
 	} else if org, err := db.GetOrganizationByID(ctx, *auth.CurrentOrgID()); err != nil {
 		return err
-	} else if org.ID.String() != name.OrgName {
+	} else if org.Slug != nil && *org.Slug != name.OrgName {
```
christophenne: shouldn't a `nil` slug also deny access to the registry in general?

kosmoz: yes I think you're right

PR #645, opened 2025-03-16
feat: docker swarm support

asherAbecasiss:
merged upstream main into branch and resolved conflicts, integrate Docker Compose and Swarm into a unified package.


frontend/ui/src/app/applications/application-detail.component.ts
```diff
@@ -111,7 +111,12 @@ export class ApplicationDetailComponent implements OnInit, OnDestroy {
       template: new FormControl(''),
     }),
     docker: new FormGroup({
-      compose: new FormControl('', Validators.required),
+      dockerType: new FormControl<DockerType>('compose', {
+        nonNullable: true,
+        validators: Validators.required,
+      }),
+      compose: new FormControl(''),
+      swarm: new FormControl(''),
```
kosmoz: I think this field is unused

internal/db/applications.go
```diff
@@ -210,14 +210,20 @@ func GetApplicationForApplicationVersionID(ctx context.Context, id, orgID uuid.U
 }
 
 func CreateApplicationVersion(ctx context.Context, applicationVersion *types.ApplicationVersion) error {
+	// defaultDockerType := DockerTypeCompose
+	// av.Docker_Type = &defaultDockerType
+	fmt.Println("----------------->>>>>>>>>>>>", applicationVersion.Docker_Type)
```
kosmoz: should probably be removed

internal/types/application_version.go
```diff
@@ -11,16 +11,16 @@ import (
...
-	ChartType    *HelmChartType `db:"chart_type" json:"chartType,omitempty"`
-	ChartName    *string        `db:"chart_name" json:"chartName,omitempty"`
-	ChartUrl     *string        `db:"chart_url" json:"chartUrl,omitempty"`
-	ChartVersion *string        `db:"chart_version" json:"chartVersion,omitempty"`
+	ID            uuid.UUID      `db:"id" json:"id"`
+	CreatedAt     time.Time      `db:"created_at" json:"createdAt"`
+	ArchivedAt    *time.Time     `db:"archived_at" json:"archivedAt,omitempty"`
+	Name          string         `db:"name" json:"name"`
+	ApplicationID uuid.UUID      `db:"application_id" json:"applicationId"`
+	Docker_Type   *DockerType    `db:"docker_type" json:"dockerType,omitempty"`
```
kosmoz: Please rename to `DockerType`

internal/types/application_version.go
```diff
@@ -60,8 +60,10 @@ func (av ApplicationVersion) ParsedComposeFile() (result map[string]any, err err
 
 func (av ApplicationVersion) Validate(deplType DeploymentType) error {
 	if deplType == DeploymentTypeDocker {
-		if av.ComposeFileData == nil {
-			return errors.New("missing compose file")
+		if av.Docker_Type == nil || *av.Docker_Type == "" {
+			return errors.New("missing docker type")
+		} else if *av.Docker_Type == DockerTypeCompose && av.ComposeFileData == nil {
```
kosmoz: Wouldn't a "swarm" application also require a compose file?

kosmoz (commented):
Hi @asherAbecasiss! At first glance this looks promising.

The one thing I'm not sure about having the `DockerType` on the application version entity. That is because, as far as I understand (please correct me if I'm wrong), `docker compose` and `docker stack` are both built on the same file specification, where both ignore some parts that are not relevant. I think it might make more sense to have the `DockerType` on the deployment target entity, that way a vendor only has to create an application once and the customer can decide which type to use later. What do you think, would that work?

internal/handlers/agent.go
```diff
@@ -177,6 +177,7 @@ func agentResourcesHandler(w http.ResponseWriter, r *http.Request) {
 				} else {
 					agentDeployment.ComposeFile = patchedComposeFile
 					agentDeployment.EnvFile = deployment.EnvFileData
+					agentDeployment.DockerType = *deployment.DockerType
```
christophenne: technically `deployment.DockerType` could be nil here, would feel safer with a nil check?

cmd/agent/docker/agent_deployment.go
```diff
@@ -8,13 +8,15 @@ import (
...
 )
 
 type AgentDeployment struct {
-	ID          uuid.UUID `json:"id"`
-	RevisionID  uuid.UUID `json:"revisionId"`
-	ProjectName string    `json:"projectName"`
+	ID          uuid.UUID        `json:"id"`
+	RevisionID  uuid.UUID        `json:"revisionId"`
+	ProjectName string           `json:"projectName"`
+	DockerType  types.DockerType `json:"docker_type,omitempty"`
```
christophenne, line 19: shouldn't it be `*types.DockerType`?

internal/handlers/agent.go
kosmoz: it's never nil for a docker deployment

cmd/agent/docker/agent_deployment.go
kosmoz: yes true

internal/handlers/agent.go
kosmoz: I used `util.PtrCopy` to resolve the other comment which includes a nil check

PR #649, opened 2025-03-17
feat: artifact license management

christophenne:


internal/db/artifacts.go
```diff
@@ -276,9 +276,14 @@ func CheckLicenseForArtifact(ctx context.Context, orgName, name, reference strin
 			SELECT av.id, av.artifact_id, av.manifest_blob_digest
 				FROM Artifact a
 				JOIN ArtifactVersion av ON a.id = av.artifact_id
-				WHERE a.organization_id = @orgName
+				JOIN Organization o ON o.id = a.organization_id
+				WHERE o.slug = @orgName
 				AND a.name = @name
-				AND (av.name = @reference OR av.manifest_blob_digest = @reference)
+				AND av.manifest_blob_digest IN (
```
christophenne: We only set the license on the "digest"-version, but not on the tagged versions. So I think its better to extend the comparison here, instead of adding the tagged version to the license too, because that way we would also have to keep updates and new tags of existing versions in sync, and that would be more difficult I think.

internal/registry/authz/authorizer.go
```diff
@@ -62,7 +62,7 @@ func (a *authorizer) AuthorizeReference(ctx context.Context, nameStr string, ref
 		return err
 	} else if org.Slug == nil || *org.Slug != name.OrgName {
 		return ErrAccessDenied
-	} else if action == ActionRead && *auth.CurrentUserRole() != types.UserRoleVendor {
+	} else if action != ActionWrite && *auth.CurrentUserRole() != types.UserRoleVendor {
```
christophenne, line 65: at least I think it should also authorize the stat action? not sure

kosmoz: Not sure, but let's keep it this way for now and see if it causes any problems :+1:

internal/db/artifacts.go
kosmoz: I wanted to comment that I think it would be better to set the license on the "tag"-version instead but the way you did it makes way more sense!

However, I find this query a bit hard to grok. I don't understand why just checking `av.name` and `av.manifest_blob_digest` isn't enough, can you explain pls?

kosmoz (commented):
One thing I don't undetstand, otherwise looking good

internal/db/artifacts.go
christophenne: The subquery is just "translate tag reference to its `manifest_blob_digest`", such that the comparison is always against the digests. Probably its a bit too complicated. 

Example: user does `docker pull registry.local/org/artifact:1.3.3`, then the `@reference` param in this query is `1.3.3`. So we need to translate this tag reference to its corresponding digest. And this digest we can check the license against.

PR #653, opened 2025-03-18
feat(registry): add support for chunked blob uploads

kosmoz:


internal/registry/blob.go
```diff
@@ -436,37 +419,43 @@ func (b *blobs) handlePut(resp http.ResponseWriter, req *http.Request, service,
...
+	if contentRange != "" {
+		start, end := 0, 0
+		if _, err := fmt.Sscanf(contentRange, "%d-%d", &start, &end); err != nil {
+			return &regError{
+				Status:  http.StatusRequestedRangeNotSatisfiable,
+				Code:    "BLOB_UPLOAD_UNKNOWN",
+				Message: "We don't understand your Content-Range",
+			}
+		}
+		size, err := bph.PutChunk(req.Context(), target, req.Body, 1)
```
christophenne: not entirely familiar with the protocol, but shouldn't `start` be used somewhere, and wouldn't that be here as the last param?

christophenne: It's used in the other upload, so apparently this `1` means to not use the chunked upload?

internal/registry/blob/s3/handler.go
```diff
@@ -131,6 +245,42 @@ func (handler *blobHandler) Delete(ctx context.Context, repo string, h v1.Hash)
...
+}
+
+func (handler *blobHandler) getExistingParts(
+	ctx context.Context,
+	uploadKey string,
+	uploadID string,
+) ([]types.Part, error) {
+	// ListParts returns at most 1000 elements.
+	// Thus, we can not currently handle uploads with more than 1000 chunks!
+	if result, err := handler.s3Client.ListParts(ctx, &s3.ListPartsInput{
```
christophenne, line 274: optional, but we could check whether the limit is reached (`result.IsTruncated`) and return an error in that case? Otherwise, how would the information be propagated back that the list might be incomplete.

kosmoz: good idea, will do that

internal/registry/blob.go
kosmoz: good point, it should probably use `start` instead

PR #660, opened 2025-03-19
feat: artifact audit logging

christophenne:


frontend/ui/src/app/services/users.service.ts
```diff
@@ -15,24 +18,72 @@ export interface CreateUserAccountResponse {
...
+}
+
 @Injectable({providedIn: 'root'})
 export class UsersService {
-  private readonly httpClient = inject(HttpClient);
   private readonly baseUrl = '/api/v1/user-accounts';
+  private readonly cache: ReactiveList<UserAccountWithRole>;
+  private readonly auth = inject(AuthService);
+
+  constructor(private readonly httpClient: HttpClient) {
```
kosmoz: I personally prefer to use `inject` over constructor injection but at least I think we should try to use only one of them per class :smile: 

The advantage of `inject` is that you could also initialize `cache` where it is declared.

frontend/ui/src/app/services/users.service.ts
```diff
@@ -15,24 +18,72 @@ export interface CreateUserAccountResponse {
...
 
   public getUserStatus(): Observable<{active: boolean}> {
     return this.httpClient.get<{active: boolean}>(`${this.baseUrl}/status`);
   }
 
   public addUser(request: CreateUserAccountRequest): Observable<CreateUserAccountResponse> {
-    return this.httpClient.post<CreateUserAccountResponse>(this.baseUrl, request);
+    return this.httpClient.post<CreateUserAccountResponse>(this.baseUrl, request).pipe(
+      tap((it) =>
+        this.cache.save({
```
kosmoz, line 59: :thinking: 
```suggestion
        // TODO: add user details to CreateUserAccountResponse
        this.cache.save({
```

frontend/ui/src/app/services/users.service.ts
```diff
@@ -15,24 +18,72 @@ export interface CreateUserAccountResponse {
...
+    );
+  }
+
+  public delete(user: UserAccountWithRole): Observable<void> {
+    return this.httpClient.delete<void>(`${this.baseUrl}/${user.id}`).pipe(tap(() => this.cache.remove(user)));
   }
 
-  public delete(user: UserAccount): Observable<void> {
-    return this.httpClient.delete<void>(`${this.baseUrl}/${user.id}`);
+  public getUserWithGravatarUrl(id: string): Observable<{user: UserAccountWithRole; gravatar: string} | undefined> {
```
kosmoz: I think `UserAccountWithRole & {gravatar: string}` would be a bit cleaner.

Also I think it would be better to throw an error in case the user is not found and handle that error downstream with `catchError(() => NEVER)` or something, rather than having to handle `undefined` everywhere.

internal/registry/manifest.go
```diff
@@ -344,11 +350,17 @@ func (handler *manifests) handleGet(resp http.ResponseWriter, req *http.Request,
 	if _, err := io.Copy(resp, &buf); err != nil {
 		return regErrInternal(err)
 	}
+	if err := handler.audit.AuditPull(ctx, repo, target); err != nil {
```
kosmoz, line 358: Bug: This is never reached when redirects are enabled (which they are by default). Please also add calling of `AuditPull` above, where the `RedirectError` is handled.

After trying this, there is another error: 
> could not query ArtifactVersion: too many rows in result set

I was able to fix this by changing `db.GetDigestArtifactVersion` to use `SELECT DISTINCT`.

kosmoz (changes requested):

frontend/ui/src/app/services/users.service.ts
christophenne: yes, good idea

PR #669, opened 2025-03-20
chore: read s3 config from environment

kosmoz:
I had to rewrite the env init function because the linter started complaining about cyclomatic complexity being too high :smiling_face_with_tear:


internal/env/env.go
```diff
@@ -45,115 +43,52 @@ func init() {
...
-	}
-
-	if value, ok := os.LookupEnv("FRONTEND_POSTHOG_UI_HOST"); ok {
-		frontendPosthogUIHost = &value
-	}
-
-	if value, ok := os.LookupEnv("USER_EMAIL_VERIFICATION_REQUIRED"); ok {
-		userEmailVerificationRequired = util.Require(strconv.ParseBool(value))
-	}
+	registryS3Config.Bucket = requireEnv("REGISTRY_S3_BUCKET")
```
christophenne, line 77: I think requiring all the registry values is a bit strict, but okay for now, since there is not "turning off" the registry yet. I think that could be a little addition in the future, for people that don't need it at all.

pmig: related to https://github.com/glasskube/distr/issues/677

PR #672, opened 2025-03-20
fix(ui): cancel deployment polling when not in use

christophenne:


frontend/ui/src/app/services/deployment-targets.service.ts
```diff
@@ -42,7 +42,10 @@ export class DeploymentTargetsService implements CrudService<DeploymentTarget> {
           ? timer(Math.min(Math.pow(c, 2), 30) * 1000)
           : EMPTY,
     }),
-    shareReplay(1)
+    shareReplay({
+      bufferSize: 1,
+      refCount: true,
```
kosmoz, line 47: Interesting, I was pretty sure refCount=true would be the default, but turns out it is not

kosmoz (approved):

PR #676, opened 2025-03-21
chore(registry): respond with 416 when uploading chunk multiple times

christophenne:


internal/registry/blob/s3/handler.go
```diff
@@ -157,7 +157,10 @@ func (handler *blobHandler) PutChunk(ctx context.Context, id string, r io.Reader
 	var size int64
 
 	if start == 0 {
-		if upload, err := handler.s3Client.CreateMultipartUpload(ctx, &s3.CreateMultipartUploadInput{
+		if _, err := handler.getUploadID(ctx, uploadKey); !errors.Is(err, blob.ErrBadUpload) {
```
kosmoz: I think the `err != nil` case should be handled separately here

EDIT: Or rather, I think the "not found" case should be handled explicitly and other errors should return a different error

christophenne: You mean like this?

kosmoz: I see now that getUploadID returns ErrBadUpload iff the upload ID doesn't exist. I think to make it less confusing something like this would be best:
```go
if err == nil {
  // upload ID must not exist if start == 0!
  return 0, blob.NewErrBadUpload("range is not as expected")
} else if !errors.Is(err, blob.ErrBadUpload) {
  return 0, err
}
```

What do you think?

christophenne: yes, will do it like that :+1:

Issue #677, opened 2025-03-21
Deploy examples don't support the registry yet


mjtechguy:
Seems the new version requires some default information regarding S3 in the .env before it will work, otherwise it panics.

```
hub-1       | panic: missing required environment variable: REGISTRY_S3_BUCKET
hub-1       | 
hub-1       | goroutine 1 [running]:
hub-1       | github.com/glasskube/distr/internal/env.requireEnv({0x1453d49, 0x12})
hub-1       |   /home/runner/work/distr/distr/internal/env/util.go:52 +0x88
hub-1       | github.com/glasskube/distr/internal/env.init.0()
hub-1       |   /home/runner/work/distr/distr/internal/env/env.go:77 +0x64e
hub-1       | panic: missing required environment variable: REGISTRY_S3_BUCKET
```

I added the following and it worked (not even valid, but the vars needed to exist).

```
REGISTRY_S3_BUCKET="distr-registry"
REGISTRY_S3_REGION="us-east-1"
REGISTRY_S3_ENDPOINT="http://minio:9000"
```

pmig:
This is indeed correct our deployment docker-compose and Helm Chart have not yet been updated. Will fix it asap.

pmig:
Hi @mjtechguy we just releases version 1.4.1 which includes registry support for the docker and Helm examples. Please let us know if everything works now as expected.

PR #684, opened 2025-03-24
chore(registry): support REGISTRY_ENABLED env variable

christophenne:
Ref #677


deploy/docker/.env
```diff
@@ -6,7 +6,17 @@ DATABASE_URL=postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/${PO
 
 # General
 DISTR_HOST="https://localhost:8080"
-DISTR_ARTIFACTS_HOST="localhost:8585"
+DISTR_ARTIFACTS_HOST="localhost:8585" # only served if REGISTRY_ENABLED is true
```
christophenne: Unfortunately, naming has become a bit inconsistent, all registry things should have either `REGISTRY` or `ARTIFACTS` in their name. 

We can decide for one and introduce fixes, but I think at this point we should only deprecate the outdated ones, then?

internal/server/servable.go
```diff
@@ -0,0 +1,9 @@
+package server
+
+import "context"
+
+type Servable interface {
```
christophenne: Such a Java name – suggestions welcome.

kosmoz: Maybe call it `Server` and rename the concrete implementation to `server`?

deploy/docker/.env
pmig: I would prefer to rename it to DISTR_REGISTRY_HOST (and _only_ / use registry in the env) and prefer the terminology registry (artifacts are uploaded to the registry)

future vars could be DISTR_REGISTRY_ARTIFACT_LIMIT

kosmoz: fine by me

.env.development.local
```diff
@@ -23,8 +22,12 @@ STATUS_ENTRIES_MAX_AGE=1h
 # SENTRY_DSN="https://ef0c6d7ca9b61d227dafefaaf3199901@o4508443344633856.ingest.de.sentry.io/4508443371569232"
 # SENTRY_DEBUG=true
 
-# REGISTRATION=enabled
+# user registration mode
+# REGISTRATION=enabled # can be one of "enabled" (default), "hidden", "disabled"
 
+# OCI Registry Settings
+REGISTRY_ENABLED=false # if true, REGISTRY_S3_BUCKET and REGISTRY_S3_REGION are mandatory
+DISTR_REGISTRY_HOST="localhost:8585" # only served if REGISTRY_ENABLED is true, defaults to DISTR_HOST
```
pmig:
```suggestion
REGISTRY_HOST="localhost:8585" # only served if REGISTRY_ENABLED is true, defaults to DISTR_HOST
```

pmig: not mandatory, but seems more fitting?

christophenne: agree :+1:

kosmoz (approved):

PR #691, opened 2025-03-25
chore(ui): show registry instructions

christophenne:


frontend/ui/src/app/artifacts/artifacts/artifacts.component.html
```diff
@@ -1,5 +1,54 @@
...
+        @if (registrySlug$ | async; as slug) {
+          @if (registryHost$ | async; as host) {
+            <div class="font-medium">
+              To use the OCI Registry, make sure to create a
+              <a class="text-gray-600 dark:text-gray-400 underline" routerLink="/settings/access-tokens"
+                >personal access token</a
+              >
+              and log in to the registry from your machine:
+              <code class="mt-1 select-all block text-sm text-gray-900 dark:text-gray-200 whitespace-pre-line">
+                echo "your-personal-access-token-here" | docker login {{ host }}
```
kosmoz: This doesn't work for me. This does:
```sh
echo "..." | docker login localhost:8585 --password-stdin -u -
```

kosmoz (commented):
Unrelated to this PR, but shouldn't `.env.development.local` have `REGISTRY_ENABLED=true`?

frontend/ui/src/app/artifacts/artifacts/artifacts.component.html
christophenne: True, thanks for checking.

kosmoz (approved):

PR #703, opened 2025-03-26
chore(registry): introduce optional tag limit per organization

christophenne:


kosmoz (commented):
Looks good to me. Just one small comment

internal/registry/error.go
```diff
@@ -112,3 +112,11 @@ var regErrDenied = &regError{
 	Code:    "DENIED",
 	Message: "Access to the resource has been denied",
 }
+
+func regErrDeniedQuotaExceeded(msg string) *regError {
```
kosmoz: I think this could be changed to a `var`.

I would also alter the message slightly to something like this: "You have exhausted your organizations tag quota"

kosmoz (approved):

PR #733, opened 2025-04-01
chore(backend): enhanced token verification and improved visibility constraints

christophenne:
* user and organization of JWTs and access tokens are loaded into the request context via the authentication chain
* `aud` key added to the generated JWTs for later use (right now the possibilities are "user" and "agent")
* improved access control checking regarding deployments

### DIFF
--- a/internal/authn/authinfo/db.go
+++ b/internal/authn/authinfo/db.go
@@ -29,20 +29,17 @@ func DbAuthenticator() authn.Authenticator[AuthInfo, AuthInfo] {
 		var user *types.UserAccount
 		var org *types.Organization
 		var err error
-		if a.CurrentOrgID() == nil {
-			if user, err = db.GetUserAccountByID(ctx, a.CurrentUserID()); errors.Is(err, apierrors.ErrNotFound) {
-				return nil, authn.ErrNoAuthentication
-			}
-		} else {
-			var userWithRole *types.UserAccountWithUserRole
-			if userWithRole, org, err = db.GetUserAccountAndOrgWithRole(
-				ctx, a.CurrentUserID(), *a.CurrentOrgID()); errors.Is(err, apierrors.ErrNotFound) {
-				return nil, authn.ErrBadAuthentication
-			} else if err != nil {
-				return nil, err
-			} else if a.CurrentUserRole() != nil && userWithRole.UserRole != *a.CurrentUserRole() {
+		if a.CurrentOrgID() != nil {
+			if user, org, err = db.GetUserAccountAndOrg(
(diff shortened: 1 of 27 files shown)


internal/authn/authinfo/db.go
```diff
@@ -0,0 +1,50 @@
...
+		var user *types.UserAccount
+		var org *types.Organization
+		var err error
+		if a.CurrentOrgID() != nil {
+			if user, org, err = db.GetUserAccountAndOrg(
+				ctx, a.CurrentUserID(), *a.CurrentOrgID(), *a.CurrentUserRole()); errors.Is(err, apierrors.ErrNotFound) {
+				return nil, authn.ErrBadAuthentication
+			}
+		} else if user, err = db.GetUserAccountByID(ctx, a.CurrentUserID()); errors.Is(err, apierrors.ErrNotFound) {
+			return nil, authn.ErrNoAuthentication
```
kosmoz: Shouldn't this be `ErrBadAuthentication`, since a user ID was present but the user doesn't exist?

internal/authn/authinfo/db.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (a DbAuthInfo) CurrentUser() *types.UserAccount {
+	return a.user
+}
+
+func (a DbAuthInfo) CurrentOrg() *types.Organization {
+	return a.org
+}
+
+func DbAuthenticator() authn.Authenticator[AuthInfo, AuthInfo] {
```
kosmoz: I think it should be possible to change this from [AuthInfo, AuthInfo] to [AuthInfo, *DBAuthInfo] without too many changes. This would allow for removing the new functions from the AuthInfo interface and the stubs in SimpleAuthInfo that just panic

internal/middleware/middleware.go
```diff
@@ -115,6 +143,23 @@ var RequireOrgID = auth.Authentication.ValidatorMiddleware(func(value authinfo.A
...
+	if value.CurrentUser() == nil {
+		return authn.ErrBadAuthentication
+	} else {
+		return nil
+	}
+})
+
+var RequireUserOrgRole = auth.Authentication.ValidatorMiddleware(func(value authinfo.AuthInfo) error {
+	if value.CurrentOrgID() == nil || value.CurrentOrg() == nil ||
+		value.CurrentUser() == nil || value.CurrentUserRole() == nil {
```
kosmoz: is it even possible for the user to be `nil`?

internal/authn/authinfo/db.go
christophenne: yes, correct

internal/middleware/middleware.go
christophenne: no not any longer

kosmoz (approved):

PR #743, opened 2025-04-03
feat: add artifact download history page

kosmoz:


frontend/ui/src/app/artifacts/artifact-versions/artifact-versions.component.ts
```diff
@@ -4,7 +4,7 @@ import {ActivatedRoute} from '@angular/router';
 import {FaIconComponent} from '@fortawesome/angular-fontawesome';
 import {faBox, faDownload, faFile, faLightbulb, faWarning} from '@fortawesome/free-solid-svg-icons';
 import dayjs from 'dayjs';
-import {distinctUntilChanged, filter, firstValueFrom, map, Observable, switchMap} from 'rxjs';
+import {distinctUntilChanged, filter, firstValueFrom, map, Observable, switchMap, tap} from 'rxjs';
```
christophenne:
```suggestion
import {distinctUntilChanged, filter, firstValueFrom, map, Observable, switchMap} from 'rxjs';
```

internal/db/artifacts.go
```diff
@@ -480,11 +481,9 @@ func GetDigestArtifactVersion(ctx context.Context, orgName, name, reference stri
 		FROM Artifact a
 		JOIN Organization o ON o.id = a.organization_id
 		JOIN ArtifactVersion v ON a.id = v.artifact_id
-		JOIN ArtifactVersion avx ON a.id = avx.artifact_id AND v.manifest_blob_digest = avx.manifest_blob_digest
 		WHERE o.slug = @orgName
 			AND a.name = @name
```
christophenne: I think this function is now the same as `GetArtifactVersion` above and can be removed.

internal/types/types.go
```diff
@@ -67,3 +67,7 @@ func (target *Digest) Scan(src any) error {
 func (src Digest) TextValue() (pgtype.Text, error) {
 	return pgtype.Text{String: v1.Hash(src).String(), Valid: true}, nil
 }
+
+func (h Digest) MarshalJSON() ([]byte, error) {
```
christophenne, line 71: what is this for?

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.html
```diff
@@ -0,0 +1,50 @@
...
+                    {{ pull.userAccount.name ?? pull.userAccount.email }}
+                  } @else {
+                    unknown
+                  }
+                </td>
+                <td class="px-4 py-3 font-medium text-gray-900 whitespace-nowrap dark:text-white">
+                  {{ pull.remoteAddress }}
+                </td>
+                <td class="px-4 py-3 font-medium text-gray-900 whitespace-nowrap dark:text-white">
+                  {{ pull.createdAt | date: 'medium' }}
```
christophenne: The date format differs from the one we usually use.

internal/db/artifacts.go
kosmoz: You're right :+1:

internal/types/types.go
kosmoz: without this, the digest is serialized like this:
```json
{"Algorithm": "sha265", "Hash": "..."}
```
But we want the string representation:
```json
"sha265:..."
```
This is needed because the `Digest` type definition does not inherit the methods of the underlying `Hash` type.

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.html
kosmoz: we seem to be using a mix already :sweat_smile: 
image
But I changed this one to `short`

kosmoz:
> Apart from that, a possible future enhancements that is out of scope for this PR (if desired, we should create issues): make the table filterable by artifact(version), so that we can link from the artifacts pages to the download history of this artifact. There are probably a lot of similar reporting/analytics cases.

Definitely needed in the future, but for now I wanted to keep it simple

internal/types/types.go
christophenne: I see, thx

PR #747, opened 2025-04-04
feat: agent registry auth

kosmoz:


internal/agentenv/env.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"strconv"
+	"time"
+
+	"github.com/glasskube/distr/internal/envutil"
+)
+
+var (
+	AgentVersionID         = envutil.GetEnv("DISTR_AGENT_VERSION_ID")
+	Interval               = envutil.GetEnvParsedOrDefault("DISTR_INTERVAL", time.ParseDuration, 5*time.Second)
```
christophenne: minor improvement: use existing `getPositiveDuration` as parsing function instead

api/interface.go
```diff
@@ -0,0 +1 @@
+package api
```
christophenne: still needed?

internal/agentauth/auth.go
```diff
@@ -34,6 +44,17 @@ func EnsureAuth(ctx context.Context, deployment api.AgentDeployment) (auth.Clien
 		}
 	}
 
+	if agentenv.DistrRegistryHost != "" {
+		if err := client.LoginWithOpts(
+			auth.WithLoginContext(ctx),
```
christophenne, line 49: It seems this is missing a `auth.WithLoginInsecure()` option in case `agentenv.DistrRegistryPlainHTTP = true`, otherwise docker login fails in the local http environment

christophenne: (also a few lines below where login is done if auth has changed)

api/interface.go
kosmoz: no :slightly_smiling_face:

internal/agentauth/auth.go
kosmoz: Thanks, I only added it here for now, since `DistrRegistryPlainHTTP` is intended to be used with the distr registry only, not other registries

kosmoz:
> Nice work, left some comments. I tested the docker agent and it works fine, but I am having trouble testing the kubernetes one in a local environment with minikube and no https (even with the alleged fix that I commented).

Can you please try again? It worked for me with the following config: 
```env
DISTR_HOST="http://host.minikube.internal:8080"
REGISTRY_HOST="host.minikube.internal:8585"
REGISTRY_S3_ENDPOINT="http://host.minikube.internal:9000"
```

(and hardcoding `DISTR_REGISTRY_PLAIN_HTTP: "true"` in the kubernetes agent mainfest template)

PR #764, opened 2025-04-09
chore: user and application icons

pmig:


internal/migrations/sql/32_images.down.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE UserAccount
```
christophenne, line 1: before merging, check migration number again (could conflict with 32+33 of the tutorials branch if its merged first)

internal/types/user_account.go
```diff
@@ -16,21 +16,23 @@ type UserAccount struct {
...
-	// copy+pasted from UserAccount because pgx does not like embedded strucs
+	// copy+pasted from UserAccount because pgx does not like embedded structs
 	ID              uuid.UUID  `db:"id" json:"id"`
 	CreatedAt       time.Time  `db:"created_at" json:"createdAt"`
 	Email           string     `db:"email" json:"email"`
 	EmailVerifiedAt *time.Time `db:"email_verified_at" json:"-"`
 	PasswordHash    []byte     `db:"password_hash" json:"-"`
 	PasswordSalt    []byte     `db:"password_salt" json:"-"`
 	Name            string     `db:"name" json:"name,omitempty"`
+	ImageID         *uuid.UUID `db:"image_id" json:"imageId,omitempty"`
```
christophenne: is there a reason why this is returned in the json and the above `UserAccount` not? if not, I would keep them same

frontend/ui/src/app/components/image-upload/image-upload-dialog.component.html
```diff
@@ -0,0 +1,71 @@
...
+        </button>
+      </div>
+      <!-- Modal body -->
+      <div class="p-4 md:p-5 space-y-4">
+        <div class="grid gap-4">
+          <div class="space-y-4">
+            <div>
+              <div>
+                @if (imageSrc | async; as image) {
+                  <img class="h-48 mb-4" [src]="image" alt="Image" />
```
christophenne, line 23: general question: shouldn't delete or "reset to default" also be an option? (but can probably be added later on)

frontend/ui/src/util/secureImage.ts
```diff
@@ -0,0 +1,21 @@
...
+  private readonly httpClient = inject(HttpClient);
+  private readonly domSanitizer = inject(DomSanitizer);
+
+  transform(url?: string): Observable<SafeUrl> {
+    if (!url || !url.length) {
+      return of('/distr-logo.svg');
+    }
+    return this.httpClient
+      .get(url, {responseType: 'blob'})
+      .pipe(map((val) => this.domSanitizer.bypassSecurityTrustUrl(URL.createObjectURL(val))));
```
christophenne: I don't think bypassing should be necessary. What was the reason for this?

frontend/ui/src/app/applications/application-detail.component.html
```diff
@@ -55,12 +55,23 @@
         <div class="my-4 gap-4 md:my-6 md:gap-12">
           <div class="space-y-4">
             <div class="flex space-x-4">
-              <img
-                [ngSrc]="'/' + application.type + '.png'"
-                [alt]="application.type"
-                class="h-16 w-16 rounded-sm"
-                height="199"
-                width="199" />
+              @if (application.imageUrl) {
```
christophenne, line 58: editing the name of the app resets the app image to docker/kube icon, because the PUT call doesn't return the image url yet

frontend/ui/src/app/components/image-upload/image-upload-dialog.component.html
```diff
@@ -0,0 +1,71 @@
...
+                    class="w-full text-sm text-gray-900 border border-gray-300 rounded-lg cursor-pointer bg-gray-50 dark:text-gray-400 focus:outline-none dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400"
+                    aria-describedby="file_input_help"
+                    id="file_input"
+                    type="file" />
+                }
+              </div>
+
+              <p class="mt-1 mb-3 text-xs font-normal text-gray-500 dark:text-gray-400" id="file_input_help">
+                SVG, PNG, JPG or GIF (recommended size 32px x 32px - square). If not set, Gravatar will be used as
+                fallback.
```
christophenne:
```suggestion
                SVG, PNG, JPG or GIF (recommended size 32px x 32px - square).
```

frontend/ui/src/app/components/image-upload/image-upload-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  private toast = inject(ToastService);
+  private files = inject(FilesService);
+
+  protected readonly form = new FormGroup({
+    image: new FormControl<Blob | null>(null, Validators.required),
+  });
+
+  formLoading = signal(false);
+  protected readonly imageSrc: Observable<string | null> = this.form.controls.image.valueChanges.pipe(
```
christophenne, line 38: please only pipe until destroyed

internal/db/user_accounts.go
```diff
@@ -321,3 +321,25 @@ func UpdateUserAccountLastLoggedIn(ctx context.Context, userID uuid.UUID) error
...
+	rows, err := db.Query(ctx,
+		`UPDATE UserAccount as u SET image_id = @imageId WHERE id = @id RETURNING `+userAccountOutputExpr,
+		pgx.NamedArgs{
+			"imageId": imageID,
+			"id":      userAccount.ID,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("could not query users: %w", err)
+	} else if updated, err := pgx.CollectExactlyOneRow[types.UserAccount](rows, pgx.RowToStructByName); err != nil {
```
christophenne: I think it would be (minimally) more stable if in such cases we don't return a whole object, but only an ID or something, because:
* if something changes in the future, parsing errors won't happen that easily
* we don't use the returned object anyway, in this case we only copy the `ImageID` (that we have already as parameter anyway)

I know its similar in other places, but for the few new ones I would change it to only parse the ID (or use another way to check that exactly one row has been updated).

api/file.go
```diff
@@ -0,0 +1,16 @@
...
+
+type PatchImageRequest struct {
+	ImageID uuid.UUID `json:"imageId"`
+}
+
+func WithImageUrl(imageID *uuid.UUID) string {
+	if imageID == nil || uuid.Nil == *imageID {
+		return ""
+	}
+	return "/api/v1/files/" + imageID.String()
```
christophenne, line 15: naive question, but why doesn't the client simply append these two strings when it needs a file?

I also have no problem with the response objects generally, but it seems a bit far fetched to introduce them for this.

christophenne: it could be simply another pipe in front of the `secureImage` pipe that does this little concatenation

internal/handlers/file.go
```diff
@@ -0,0 +1,143 @@
...
+	auth := auth.Authentication.Require(ctx)
+
+	if file, err := getFileFromRequest(r); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	} else if err := db.CreateFile(ctx, *auth.CurrentOrgID(), file); err != nil {
+		log.Warn("error uploading file", zap.Error(err))
+		if errors.Is(err, apierrors.ErrNotFound) {
+			w.WriteHeader(http.StatusNoContent)
+		} else if errors.Is(err, apierrors.ErrConflict) {
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
christophenne:
```suggestion
			http.Error(w, "file is still in use", http.StatusBadRequest)
```

or something like that

internal/handlers/applications.go
```diff
@@ -387,6 +389,37 @@ func deleteApplication(w http.ResponseWriter, r *http.Request) {
...
 
+func patchImageApplicationHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := internalctx.GetLogger(ctx)
+	application := internalctx.GetApplication(ctx)
+
+	body, err := JsonBody[api.PatchImageRequest](w, r)
+
+	if err != nil {
+		sentry.GetHubFromContext(ctx).CaptureException(err)
```
christophenne: Usually we don't track such bad request errors in sentry

frontend/ui/src/util/secureImage.ts
pmig: No particular reason

api/file.go
pmig: As discussed yesterday, we could directly expose the `imageID` but we may want to load images from object storage in the future, so we don't need to introduce the image url later

frontend/ui/src/app/components/image-upload/image-upload-dialog.component.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.destroyed$.next();
+    this.destroyed$.complete();
+  }
+
+  async save() {
+    this.form.markAllAsTouched();
+    if (this.form.valid) {
+      this.formLoading.set(true);
+      const formData = new FormData();
+      formData.set('file', this.form.value.image ? (this.form.value.image as File) : '');
```
kosmoz: Is there a valid case where `this.form.value.image` is "falsy"? Since the form control has the "required" validator, I assume no. So, I think it would be simpler to just assert that it isn't.
```suggestion
      formData.set('file', this.form.value.image as File);
```

api/application.go
```diff
@@ -0,0 +1,35 @@
+package api
+
+import "github.com/glasskube/distr/internal/types"
+
+type ApplicationResponse struct {
+	types.Application
+	ImageUrl string `json:"imageUrl"`
+}
+
+func AsApplication(a *types.Application) ApplicationResponse {
```
kosmoz: I don't think this function should accept a pointer. A pointer makes sense in two cases: 
1. the function modifies it's argument
2. the argument can be nil

I don't think either applies here.

Same goes for the other mapping functions.

frontend/ui/src/app/components/users/users.component.ts
```diff
@@ -139,6 +143,14 @@ export class UsersComponent implements OnDestroy {
     }
   }
 
+  public async uploadImage(data: UserAccountWithRole) {
+    const fileId = await firstValueFrom(this.overlay.uploadImage({imageUrl: data.imageUrl}));
+    if (!fileId || data.imageUrl?.includes(fileId)) {
+      return;
+    }
+    await firstValueFrom(this.users.patchImage(data.id!!, fileId));
```
kosmoz: The second `!` doesn't do anything (I think you might have mixed that up with Kotlin syntax)
```suggestion
    await firstValueFrom(this.users.patchImage(data.id!, fileId));
```

frontend/ui/src/app/artifacts/artifact-versions/artifact-versions.component.ts
```diff
@@ -149,6 +153,11 @@ export class ArtifactVersionsComponent {
...
   }
 
-  protected readonly faWarning = faWarning;
-  protected readonly faFile = faFile;
+  public async uploadImage(data: ArtifactWithTags) {
+    const fileId = await firstValueFrom(this.overlay.uploadImage({imageUrl: data.imageUrl}));
+    if (!fileId || data.imageUrl?.includes(fileId)) {
+      return;
+    }
+    await firstValueFrom(this.artifacts.patchImage(data.id!!, fileId));
```
kosmoz:
```suggestion
    await firstValueFrom(this.artifacts.patchImage(data.id!, fileId));
```

frontend/ui/src/app/applications/application-detail.component.ts
```diff
@@ -346,5 +353,11 @@ export class ApplicationDetailComponent implements OnInit, OnDestroy {
     }
   }
 
-  protected readonly faMagnifyingGlass = faMagnifyingGlass;
+  public async uploadImage(data: Application) {
+    const fileId = await firstValueFrom(this.overlay.uploadImage({imageUrl: data.imageUrl}));
+    if (!fileId || data.imageUrl?.includes(fileId)) {
+      return;
+    }
+    await firstValueFrom(this.applicationService.patchImage(data.id!!, fileId));
```
kosmoz:
```suggestion
    await firstValueFrom(this.applicationService.patchImage(data.id!, fileId));
```

internal/types/user_account.go
```diff
@@ -16,21 +16,23 @@ type UserAccount struct {
...
 	Email           string     `db:"email" json:"email"`
 	EmailVerifiedAt *time.Time `db:"email_verified_at" json:"-"`
 	PasswordHash    []byte     `db:"password_hash" json:"-"`
 	PasswordSalt    []byte     `db:"password_salt" json:"-"`
 	Name            string     `db:"name" json:"name,omitempty"`
+	ImageID         *uuid.UUID `db:"image_id" json:"-"`
 	UserRole        UserRole   `db:"user_role" json:"userRole"` // not copy+pasted
 	Password        string     `db:"-" json:"-"`
-	// Don't forget to update AsUserAccount when adding fields!
+	// Remember to update AsUserAccount when adding fields!
```
kosmoz, line 35: You updated the comment but didn't follow it's instructions :slightly_smiling_face:

kosmoz (commented):
* application image upload says "recommended size 32x32" but it is shown with size 64x64
* the navbar still shows the gravatar (I thought these were removed?) even after changing the user avatar
* I found it a bit hard to find out how to trigger the upload dialog especially because it sometimes can be done from the list view and sometimes not. would it make sense to at least be consistent and always allow to do it from the list view? Additionally maybe add a hint on the details pages where it is possible to change an image. because otherwise I'm afraid most users will never discover this feature…
* also some small code style nitpickings

kosmoz (approved):

PR #775, opened 2025-04-10
chore(agent): kubernetes agent client config reload

kosmoz:


cmd/agent/kubernetes/main.go
```diff
@@ -245,3 +272,49 @@ func pushErrorStatus(ctx context.Context, deployment api.KubernetesAgentDeployme
...
+		}
+	}
+	for {
+		select {
+		case err, ok := <-w.Errors:
+			if !ok {
+				return nil
+			}
+			return err
+		case event := <-w.Events:
```
christophenne: In the fsnotify examples (e.g. https://github.com/fsnotify/fsnotify/blob/main/README.md#usage ) they also check the second returned flag of `.Events` (`if !ok`), I don't know if that would be appropriate here too.

cmd/agent/kubernetes/main.go
```diff
@@ -245,3 +272,49 @@ func pushErrorStatus(ctx context.Context, deployment api.KubernetesAgentDeployme
...
+			for _, dir := range dirs {
+				logger := logger.With(zap.String("dir", dir))
+				entries, err := os.ReadDir(dir)
+				if err != nil {
+					logger.Warn("read dir failed", zap.Error(err))
+					continue
+				}
+				for _, e := range entries {
+					logger := logger.With(zap.String("entry", e.Name()))
+					if e.IsDir() {
```
christophenne, line 335: I'm not sure if this works, I got some warning logs:
```
2025-04-16T05:51:34.747Z	WARN	kubernetes/main.go:311	could not update config param	{"dir": "/opt/config/.agent/env", "entry": "..data", "error": "read /opt/config/.agent/env/..data: is a directory"}
main.watchConfigDirs
	/workspace/cmd/agent/kubernetes/main.go:311
main.main.func2
	/workspace/cmd/agent/kubernetes/main.go:66

```

kosmoz: Very strange indeed, but removing the check just leads to more warnings because it actually does work for some directories … so I would just leave as-is and accept the warning, at least for now.

PR #782, opened 2025-04-14
chore(ui): disable archived versions in license edit form

christophenne:


frontend/ui/src/app/licenses/edit-license.component.html
```diff
@@ -94,19 +94,53 @@
...
+                          @for (av of archived; track av.id; let i = $index) {
+                            <li class="w-full border-gray-200 dark:border-gray-600">
+                              <div class="flex items-center">
+                                <input
+                                  type="checkbox"
+                                  id="archived-checkbox-{{ i }}"
+                                  checked
+                                  disabled
+                                  class="w-4 h-4 text-blue-600 bg-gray-100 border-gray-300 rounded-sm focus:ring-blue-500 dark:focus:ring-blue-600 dark:ring-offset-gray-700 dark:focus:ring-offset-gray-700 focus:ring-2 dark:bg-gray-600 dark:border-gray-500" />
+                                <label
```
kosmoz, line 136: To further indicate that this checkbox is disabled, I would make the text a little more muted, maybe copypasta from here: https://flowbite.com/docs/forms/checkbox/#disabled-state

kosmoz (commented):
I think it looks good, but I have one concern: Although it doesn't really make a difference, I can imagine a user trying to "unselect" an archived version in the dropdown. Would it make sense to allow this? I was able to removed archived versions from a license by changing it to unrestricted so alternatively, we could also add a title or hint for this.

kosmoz (approved):

PR #788, opened 2025-04-16
chore(ui): don't display push instructions for non vendors

pmig:


frontend/ui/src/app/artifacts/artifacts/artifacts.component.html
```diff
@@ -14,29 +14,38 @@
...
-              Then you can start tagging and pushing your images. For example, you can try the following commands to
-              push your own
-              <a
-                href="https://hub.docker.com/_/hello-world"
-                target="_blank"
-                class="text-gray-600 dark:text-gray-400 underline"
-                >hello-world</a
-              >
-              image:
+            @if (isVendor) {
```
kosmoz: I think it would be less code to use `*appRequiredRole="'vendor'"` for this

PR #795, opened 2025-04-17
chore: remove registry feature flag and reorg sidebar

christophenne:


frontend/ui/src/app/components/side-bar/side-bar.component.html
```diff
@@ -34,254 +34,195 @@
...
-              class="ml-2 bg-gray-100 text-gray-800 text-xs font-medium me-2 px-2.5 py-0.5 rounded dark:bg-gray-700 dark:text-gray-300">
-              Preview
-            </span>
-            <fa-icon [icon]="faChevronDown"></fa-icon>
-          </button>
-          @if (artifactsSubMenuOpen()) {
-            <ul class="py-2 space-y-2">
-              <li>
+          <span class="flex-1 ms-3 text-left rtl:text-right whitespace-nowrap">Agents</span>
+          <fa-icon [icon]="faChevronDown"></fa-icon>
```
kosmoz: Kind of not that important but I think it should be relatively easy to rotate the chevron 180 degrees when the menu is open. something like `[class.rotate-180]="agentsSubmenuOpen()"`

and animated with `class="transition duration-300"` (or something like that) :slightly_smiling_face:

christophenne: Good idea, added :)

PR #799, opened 2025-04-17
chore(deps): upgrade tailwindcss to v4 and flowbite to v3

kosmoz:


pmig (changes requested):
It seems that the border changes color from:

#E5E7EB to not being defined and defaulting to #000

| current | this pr |
| -- | -- |
| FireShot Capture 086 - Distr - localhost | FireShot Capture 085 - Distr - localhost |

pmig (approved):

PR #801, opened 2025-04-18
chore: add OpenTelemetry tracing

kosmoz:


cmd/hub/main.go
```diff
@@ -30,9 +30,11 @@ func main() {
 	ctx := context.Background()
 
 	util.Must(sentry.Init(sentry.ClientOptions{
-		Dsn:     env.SentryDSN(),
-		Debug:   env.SentryDebug(),
-		Release: buildconfig.Version(),
+		Dsn:              env.SentryDSN(),
+		Debug:            env.SentryDebug(),
+		EnableTracing:    true,
+		TracesSampleRate: 1.0,
```
christophenne, line 36: shouldn't the env variables be respected here too?

kosmoz: Good point, I added it for `EnableTracing`. 

For `TraceSampleRate` I'm afraid it would lead to sampling twice because the OTEL trace provider also does sampling. So I kept it at 1.0 here. But to be honest I did not test this specifically in detail.

PR #809, opened 2025-04-22
feat: add support for multiple deployments per deployment target

kosmoz:


frontend/ui/src/app/deployments/deployment-modal.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  protected readonly deployForm = new FormControl<DeploymentFormValue | undefined>(undefined, Validators.required);
+  protected readonly loading = signal(false);
+
+  protected readonly faShip = faShip;
+  protected readonly faCircleExclamation = faCircleExclamation;
+
+  constructor() {
+    effect(() => {
+      const deployment = this.deployment();
+      console.log('effect called', {deployment});
```
christophenne:
```suggestion
```

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.html
```diff
@@ -0,0 +1,583 @@
...
+                            [disabled]="!isUndeploySupported()">
+                            <fa-icon [icon]="faTriangleExclamation" class="inline-block w-4 mr-2"></fa-icon>
+                            Undeploy
+                          </button>
+                          <ng-template #undeployConfirmTemplate>
+                            <div
+                              class="p-4 text-sm text-yellow-800 rounded-lg bg-yellow-50 dark:bg-gray-800 dark:text-yellow-300"
+                              role="alert">
+                              <span class="font-semibold me-1">WARNING!</span>
+                              You are about to uninstall the application installed on {{ deploymentTarget().name }}.
```
christophenne: I think it would make more sense to put the application and version which is going to be uninstalled.

internal/db/deployment_targets.go
```diff
@@ -322,13 +322,17 @@ func CleanupDeploymentTargetStatus(ctx context.Context, dt *types.DeploymentTarg
...
-	if latest, err := GetLatestDeploymentForDeploymentTarget(ctx, dt.ID); errors.Is(err, apierrors.ErrNotFound) {
+func addDeploymentsToTarget(ctx context.Context, dt *types.DeploymentTargetWithCreatedBy) error {
+	if d, err := GetDeploymentsForDeploymentTarget(ctx, dt.ID); errors.Is(err, apierrors.ErrNotFound) {
 		return nil
 	} else if err != nil {
 		return err
 	} else {
-		dt.Deployment = latest
+		dt.Deployments = d
+		// Set the Deployment property to the first (i.e. oldest) deployment for backwards compatibility
```
christophenne, line 332: Probably makes no difference (?), but it's not always the oldest, some queries are ordered by name I think, e.g. the `func GetDeploymentTargets(ctx context.Context, orgID, userID uuid.UUID, userRole types.UserRole) ([]types.DeploymentTargetWithCreatedBy, error)`

kosmoz: But in that function we query the `DeploymentTarget` table, I think that's different… Or I don't quite understand what you mean

PR #810, opened 2025-04-22
feat: agents tutorial

christophenne:


frontend/ui/src/app/tutorials/agents/agents-tutorial.component.html
```diff
@@ -0,0 +1,264 @@
...
+                      If you need any help or in case you have questions, don't hesitate to ask in
+                      <a
+                        href="https://discord.gg/6qqBSAWZfW"
+                        target="_blank"
+                        class="text-gray-600 dark:text-gray-400 hover:underline"
+                        >Discord</a
+                      >!
+                    </p>
+                    <p class="mt-1 text-xs">
+                      When you are done, you can show us by dropping your hello-distr fork in the text field below:
```
kosmoz: Maybe change to "URL of your hello-distr fork"? I think "dropping" a fork could be confusing

internal/handlers/tutorials.go
```diff
@@ -60,18 +67,144 @@ func saveTutorialProgress(w http.ResponseWriter, r *http.Request) {
...
+		dt.AgentVersionID = &agentVersion.ID
+		if err := db.CreateDeploymentTarget(ctx, &dt, *auth.CurrentOrgID(), auth.CurrentUserID()); err != nil {
+			return nil, err
+		} else {
+			return &dt, nil
+		}
+	}
+}
+
+var helloDistrEnvironment = `
```
kosmoz: could be const I think

internal/handlers/tutorials.go
```diff
@@ -60,18 +67,144 @@ func saveTutorialProgress(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return nil, nil
+}
+
+func createHelloDistrApp(ctx context.Context) (*types.Application, error) {
+	log := internalctx.GetLogger(ctx)
+	auth := auth.Authentication.Require(ctx)
+
+	application := types.Application{
+		Name: "hello-distr",
```
kosmoz, line 134: Unrelated to this PR but I did the tutorial twice and had two apps with the same name. Would it in general make sense to require unique application names per organization? :thinking:

kosmoz (commented):
Looks good to me. I just noticed some small things. 

One more thing regarding the tutorials feature in general:
I think it would be helpful for the user if there was already a small description of each tutorial on the "list" page. That way they already get a feeling about what they will do in the tutorial and why it could be interesting for them. What do you think?

internal/handlers/tutorials.go
christophenne: Unique name per org sounds like a good idea to me (although its sort of a breaking change). Created an issue for the next triage.

kosmoz (approved):

PR #814, opened 2025-04-23
feat: add tutorials, adapt feature flag semantics and sidebar

christophenne:


pmig (changes requested):
Overall this PR is already huge and a huge improvement.  :tada: 

Following (small) changes would improve the the tutorial significantly:

 - [x] Make the Headings of a section a label for the Checkbox
 - [x] The checkbox is above the heading
 - [x] "Manual verify tasks" should be actual links with a query parameter that injects a banner to go back to the tutorial
 - [ ] Each tutorial step should be part of the url (so if you reload the page, you end up on the same tutorial)
 - [ ] Copy & Paste Icons miss the `faClipboardCheck` (and back transition)
 - [ ] For the verify deployment part I would add 2 more steps: 1. verify everything is running with `docker ps` --> go to the deployments page (so if there is an error the user sees that is is on the docker side and not on the distr side) and 2. Open Hello Distr in the Browser so the user actually has a success moment. (we can even add the cmd for the curl request to change the state of hello distr)
 - [x] Although its different from the Mockup, we redecided to want both licenses under the sidebar "group" Licenses (**Beta**) --> (1) Application Licenses (2) Artifact Licenses

frontend/ui/src/app/tutorials/agents/agents-tutorial.component.html
```diff
@@ -0,0 +1,330 @@
...
+              <form [formGroup]="deployFormGroup" (ngSubmit)="continueFromDeploy()">
+                <ol
+                  class="ml-4 relative text-gray-900 dark:text-white mb-12 border-s border-gray-200 dark:border-gray-700">
+                  <li class="mb-10 ms-8">
+                    <span
+                      class="absolute flex items-center justify-center w-8 h-8 bg-gray-100 rounded-full -start-4 ring-4 ring-white dark:ring-gray-900 dark:bg-gray-700">
+                      <fa-icon [icon]="faLightbulb"></fa-icon>
+                    </span>
+                    <h3 class="text-lg font-medium leading-tight">Deploy hello-distr</h3>
+                    <div
```
kosmoz: This looks off in dark mode. Maybe we can change it to [this one](https://flowbite.com/docs/components/alerts/#bordered-alerts)?

Also maybe only show it on MacOS (can use the [CDK platform](https://material.angular.io/cdk/platform/overview) for this)

frontend/ui/src/app/tutorials/branding/branding-tutorial.component.html
```diff
@@ -0,0 +1,254 @@
...
+                    </span>
+                    <label for="inviteDone" class="text-lg font-medium leading-tight"
+                      >Send the customer invite mail (to yourself)</label
+                    >
+                    <p class="mt-1 text-xs">
+                      For the purpose of this tutorial, make sure to send this mail to yourself.
+                    </p>
+                    <p class="mt-1 text-xs">
+                      You can either choose another one of your email addresses, but a lot of providers also support
+                      addresses in the form of username-customer&commat;provider.tld, i.e. you might be able to use
```
kosmoz: I think it should be `username+customer` instead of `username-customer`

frontend/ui/src/app/components/side-bar/side-bar.component.ts
```diff
@@ -55,10 +59,15 @@ export class SideBarComponent {
...
+  protected readonly faHome = faHome;
+  protected readonly faChevronDown = faChevronDown;
 
   @ViewChild('asideElement') private asideElement?: ElementRef<HTMLElement>;
-
-  protected readonly artifactsSubMenuOpen = signal(true);
+  protected readonly agentsSubMenuOpen = signal(true);
+  protected readonly licenseSubMenuOpen = signal(true);
+  protected readonly registrySubMenuOpen = signal(true);
+  protected readonly licenseOverlayOpen = signal(false);
```
kosmoz, line 70: Maybe something for later, but would it make sense to only have 1-2 submenus open initially and maybe persist the state in local/session storage?

internal/migrations/sql/33_tutorials.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TYPE TUTORIAL AS ENUM ('branding', 'agents', 'registry');
+
+CREATE TABLE UserAccount_Tutorial (
```
kosmoz: The Go type for this table is called TutorialProgress. Would it make sense to use the same name here? Because I wasn't initially sure if they are the same type but I think they are (?)

kosmoz (approved):
Looks good! I found some small things… nothing blocking IMO, but maybe you can take a look still?

frontend/ui/src/app/tutorials/agents/agents-tutorial.component.html
christophenne: It looks weird because of the same background:
Screenshot from 2025-04-30 11-57-39

frontend/ui/src/app/components/side-bar/side-bar.component.ts
christophenne: Sure, created an issue.

frontend/ui/src/app/tutorials/agents/agents-tutorial.component.html
kosmoz: I see where you're coming from but IMO it's still better :sweat_smile:

kosmoz (dismissed):
Looks good to me! 

Maybe we should change the PR title such that the tutorials come first in the release notes?

pmig (approved):

PR #834, opened 2025-04-29
chore: tutorials refinement

christophenne:


frontend/ui/src/app/components/side-bar/side-bar.component.html
```diff
@@ -148,43 +112,75 @@
...
+
+      <li>
+        @if (featureFlags.isLicensingEnabled$ | async) {
+          <button
+            type="button"
+            (click)="toggle(licenseSubMenuOpen)"
+            class="flex items-center w-full p-2 text-base text-gray-900 transition duration-75 rounded-lg group hover:bg-gray-100 dark:text-white dark:hover:bg-gray-700"
+            aria-controls="dropdown-example">
+            <fa-icon
+              [icon]="faUserCheck"
```
pmig:
```suggestion
              [icon]="faKey"
```

pmig (changes requested):
Mostly fine!

For consistency we should move the urls from `/licenses` and `/artifact-licenses` to `/licenses/applications` `/licenses/artifacts` so we can but the canActivate on the root root.
Also for more consistency I  would rename "Filter license" --> "Filter application licenses" (and similar)

pmig (approved):

PR #840, opened 2025-04-29
chore: new design for vendor dashboard

christophenne:


pmig (dismissed):
Looks so great! <3

We don't need to do this now but we should (now or create tickets for later):

- [x] rename the `dashboard-placeholder` to `dashboard`
- [ ] there is no need to lazy load the Home / Dashboard component any longer, but we can think about lazy loading all the children of the `NavShellComponent`, So Login / Register / Verify loads suuuper quick & light and once the user is logged in all the rest is loaded at once

pmig (dismissed):

frontend/ui/src/app/artifacts/artifacts-by-customer-card/artifacts-by-customer-card.component.html
```diff
@@ -0,0 +1,68 @@
+<div
+  class="h-full rounded-lg border border-gray-200 bg-white shadow-sm dark:border-gray-700 dark:bg-gray-800 font-medium text-gray-900 dark:text-white flex flex-col">
+  <div class="p-4 justify-between items-stretch md:flex">
+    <div class="justify-between sm:flex sm:items-center">
+      <div class="flex gap-3 items-start lg:items-center mb-2 md:mb-0">
+        <div class="relative">
+          @if (customer().imageUrl) {
+            <img class="w-8 h-8 rounded-full" [attr.src]="customer().imageUrl | secureImage | async" alt="" />
+          } @else {
+            <fa-icon [icon]="faUserCircle" size="xl" class="text-xl"></fa-icon>
```
kosmoz: This is more similar to the style used on the users list page and I think it makes the icon stand out a little less:
```suggestion
            <fa-icon [icon]="faUserCircle" size="xl" class="text-xl text-gray-400"></fa-icon>
```

frontend/ui/src/app/artifacts/artifacts-by-customer-card/artifacts-by-customer-card.component.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  protected isOnLatest(artifact: DashboardArtifact): boolean {
+    const max = this.findMaxVersion(artifact.artifact.versions ?? []);
+    const includesPulled = max?.tags.map((t) => t.name).includes(artifact.latestPulledVersion) ?? false;
+    if (includesPulled) {
+      return true;
+    }
+    const pulledStillAvailable = (artifact.artifact.versions ?? []).find((v) =>
+      v.tags.find((t) => t.name === artifact.latestPulledVersion)
+    );
```
kosmoz: both works obviously but I think `some` is semantically better fitting than `find`
```suggestion
    const pulledStillAvailable = (artifact.artifact.versions ?? []).some((v) =>
      v.tags.some((t) => t.name === artifact.latestPulledVersion)
    );
```

frontend/ui/src/app/artifacts/artifacts-by-customer-card/artifacts-by-customer-card.component.ts
```diff
@@ -0,0 +1,80 @@
...
+        versions,
+        (version) => this.getFirstSemverTag(version),
+        (a, b) => a.compare(b) > 0
+      );
+    } catch (e) {
+      return maxBy(versions, (version) => new Date(version.createdAt!));
+    }
+  }
+
+  private getFirstSemverTag(version: TaggedArtifactVersion): SemVer {
```
kosmoz: shouldn't we also take the highest one here? 

Obviously an edge case but consider something like `1.0.0-rc.0` and `1.0.0` on the same digest. It would return the RC tag bug we would want the final release.

frontend/ui/src/app/components/dashboard/dashboard.component.html
```diff
@@ -0,0 +1,39 @@
...
+      <div
+        class="p-4 mb-4 text-sm text-blue-800 max-w-128 mx-auto rounded-lg bg-blue-50 dark:bg-gray-800 dark:text-blue-400"
+        role="alert">
+        Click <a class="cursor-pointer underline" (click)="openWizard()">here</a> to continue with the onboarding
+        wizard.
+      </div>
+    }
+
+    <div class="flex flex-col gap-4">
+      <h1 class="text-3xl font-extrabold leading-none tracking-tight text-gray-900 md:text-4xl dark:text-white">
```
kosmoz: should be `h2` I think: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page

kosmoz: IMO the the headings are a little too big. Maybe `3xl`? @pmig what do you think?

internal/db/dashboard.go
```diff
@@ -0,0 +1,36 @@
...
+func GetLatestPullOfArtifactByUser(ctx context.Context, artifactId uuid.UUID, userId uuid.UUID) (string, error) {
+	db := internalctx.GetDb(ctx)
+	if rows, err := db.Query(ctx, `
+		select av.name
+		from ArtifactVersionPull avpl
+		join ArtifactVersion av on av.id = avpl.artifact_version_id
+		where av.artifact_id = @artifactId and avpl.useraccount_id = @userId
+		and av.name not like '%:%'
+		order by avpl.created_at desc
+		limit 1;
```
kosmoz: reformatted
```suggestion
		SELECT av.name
		FROM ArtifactVersionPull avpl
		JOIN ArtifactVersion av ON av.id = avpl.artifact_version_id
		WHERE av.artifact_id = @artifactId 
			AND avpl.useraccount_id = @userId
			AND av.name NOT LIKE '%:%'
		ORDER BY avpl.created_at DESC
		LIMIT 1;
```

internal/handlers/dashboard.go
```diff
@@ -0,0 +1,83 @@
...
+			customerRes := api.ArtifactsByCustomer{Customer: api.AsUserAccount(customer)}
+			for _, artifact := range artifacts {
+				if slices.Contains(artifact.DownloadedByUsers, customer.ID) {
+					if latestPulled, err := db.GetLatestPullOfArtifactByUser(ctx, artifact.ID, customer.ID); err != nil {
+						if errors.Is(err, apierrors.ErrNotFound) {
+							continue
+						} else {
+							log.Error("failed to get latest artifact pull by user", zap.Error(err),
+								zap.Any("artifactId", artifact.ID), zap.Any("userId", customer.ID))
+							sentry.GetHubFromContext(ctx).CaptureException(err)
```
kosmoz, line 55: missing return? or was it omitted on purpose?

internal/handlers/metrics.go
```diff
@@ -1,43 +0,0 @@
-package handlers
```
kosmoz, line 1: This is technically a breaking change… Although I wouldn't expect anyone to use it apart from the old dashboard, maybe we should consider having a separate "internal" api for stuff that's really only relevant for our UI where we can easily make changes like this. Can be considered for future features I guess…

package.json
```diff
@@ -46,12 +46,9 @@
     "@glasskube/distr-sdk": "file:./sdk/js/src",
     "@pandatix/js-cvss": "^0.4.4",
     "@sentry/angular": "^9.0.0",
-    "apexcharts": "^4.1.0",
     "dayjs": "^1.11.13",
     "flowbite": "^3.0.0",
-    "globe.gl": "^2.33.2",
     "jwt-decode": "^4.0.0",
-    "ng-apexcharts": "^1.15.0",
```
kosmoz, line 54: :heart:

sdk/js/docs/interfaces/UptimeMetric.md
```diff
@@ -4,22 +4,26 @@
 
 [@glasskube/distr-sdk](../README.md) / UptimeMetric
 
-# Interface: UptimeMetric
+# Interface: ~~UptimeMetric~~
```
kosmoz: Does it make sense to keep the type when the endpoint doesn't exist anymore?

kosmoz (commented):
nice work!

sdk/js/docs/interfaces/UptimeMetric.md
christophenne: Not really, will remove it.

internal/handlers/dashboard.go
christophenne: On purpose, because I wasn't sure if it can happen if somebody pulls only a digest version of an artifact? Then it would break the entire dashboard.

frontend/ui/src/app/components/dashboard/dashboard.component.html
pmig: I think they are fine (at least on my resolution - not sure about smaller devices and mobile)

internal/handlers/dashboard.go
kosmoz: OK fine then :+1:

frontend/ui/src/app/components/dashboard/dashboard.component.html
christophenne: Interesting, thanks

kosmoz (approved):

kosmoz (approved):

PR #867, opened 2025-05-05
fix(ui): improve markdown rendering on customer "home" page

kosmoz:


frontend/ui/src/app/services/markdown-options.factory.ts
```diff
@@ -1,29 +1,14 @@
...
-          <code class="text-gray-900 dark:text-gray-200 whitespace-pre-line">
-              ${code.text}
-          </code>
-        </pre>`;
-  };
+  const opts: MarkedOptions = {renderer: renderer};
 
   renderer.link = (link) => {
-    return `<a href="${link.href}" class="text-primary-600 underline" target="_blank">${link.text}</a>`;
+    const text = parseInline(link.text, {...opts, async: false});
```
christophenne: There is an endless recursion for "non-marked" links like this one:

```
The full SDK documentation is at https://distr.sh/docs/integrations/sdk/
```

Maybe it should not be called if `link.text === link.href`? But I'm not sure.

kosmoz: interesting!!

kosmoz: also added try catch to be safe

kosmoz: btw this part is needed for the "inline code inside links" style

christophenne: ah okay :+1:

PR #881, opened 2025-05-07
feat: add support for container logs

kosmoz:


.scratch/deployments/f6b498be-a2d7-48f6-9fa0-ff3e80cc7c6a
```diff
@@ -0,0 +1 @@
+{"id":"f6b498be-a2d7-48f6-9fa0-ff3e80cc7c6a","revisionId":"fb3e0293-a782-4088-a50c-ec43bee8f03d","projectName":"distr-f6b498be","docker_type":"compose","logsEnabled":true}
```
christophenne: this can be removed I guess

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.html
```diff
@@ -322,6 +322,27 @@ <h3 class="font-bold break-all">{{ deploymentTarget().name }}</h3>
...
+                        <li>
+                          <button
+                            type="button"
+                            class="text-start block w-full px-4 py-2 text-sm text-gray-700 hover:bg-gray-100 dark:text-gray-300 dark:hover:bg-gray-600 dark:hover:text-white"
+                            (click)="
+                              setLogsEnabled(deployment, !deployment.logsEnabled);
+                              showDeploymentDropdownForId.set(undefined)
+                            "
+                            [title]="
+                              isUndeploySupported()
```
christophenne: should be `isLoggingSupported()` I think

internal/agentclient/client.go
```diff
@@ -223,6 +237,8 @@ func (c *Client) ReloadFromEnv() (changed bool, err error) {
 		return
 	} else if d.metricsEndpoint, err = readEnvVar("DISTR_METRICS_ENDPOINT"); err != nil {
 		return
+	} else if d.logsEndpoint, err = readEnvVar("DISTR_LOGS_ENDPOINT"); err != nil {
```
christophenne, line 240: `DISTR_LOGS_ENDPOINT` is not being returned by the backend yet -> panic

internal/db/deployment_log_record.go
```diff
@@ -0,0 +1,133 @@
...
+		pgx.NamedArgs{"deploymentId": deploymentID},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("could not query DeploymentLogRecord: %w", err)
+	}
+	result, err := pgx.CollectRows(rows, pgx.RowTo[string])
+	if err != nil {
+		return nil, fmt.Errorf("could not collect DeploymentLogRecord: %w", err)
+	}
+	slices.Sort(result)
```
christophenne, line 53: is it out of SQL on purpose? to avoid potential un-optimized slow queries?

internal/handlers/agent.go
```diff
@@ -244,6 +247,45 @@ func agentResourcesHandler(w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+			return
+		}
+		errIdx := slices.IndexFunc(records, func(r api.DeploymentLogRecord) bool {
+			return !slices.ContainsFunc(deployments, func(d types.DeploymentWithLatestRevision) bool {
+				return d.ID == r.DeploymentID
+			})
+		})
+		if errIdx >= 0 {
+			// TODO: also check DeplyomentRevisionID
```
christophenne, line 272: still relevant? what are these checks for – to avoid an error on insert, if a deployment has been deleted meanwhile?

internal/handlers/agent.go
```diff
@@ -244,6 +247,45 @@ func agentResourcesHandler(w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, fmt.Sprintf("invalid deploymentId at index %v", errIdx), http.StatusBadRequest)
+			return
+		}
+		if err := db.SaveDeploymentLogRecords(ctx, records); err != nil {
+			log.Error("error saving log records", zap.Error(err))
```
christophenne, line 277: I think sending errors to sentry would be nice here too

internal/db/deployment_log_record.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	db := internalctx.GetDb(ctx)
+	rows, err := db.Query(
+		ctx,
+		`SELECT `+deploymentLogRecordOutputExpr+`
+		FROM DeploymentLogRecord lr
+		WHERE lr.deployment_id = @deploymentId
+			AND lr.resource = @resource
+			AND lr.timestamp < @before
+		ORDER BY lr.timestamp DESC
```
christophenne, line 76: It's probably ordered descending to be consistent with the status history, but looking at the logs it feelds weird that the more recent entries are on the top. Would it make sense to either break consistency here, or (preferrably) also make the status table ascending?

frontend/ui/src/app/deployments/deployment-status-modal/deployment-logs-table.component.ts
```diff
@@ -0,0 +1,71 @@
...
+              });
+            })
+          )
+        : of([])
+    ),
+    map((logs) => logs.map((it) => ({date: it.timestamp, status: it.severity, detail: it.body})))
+  );
+
+  constructor() {
+    effect(() => {
```
christophenne: Not sure about this, because effects should not be used to propagate state changes [1], and also it looks unintuitive. Shouldn't the same logic be possible by setting `currentOldest` and `hasMore` in a `tap` operator of `toObservable(this.resource).pipe`?

[1] https://angular.dev/guide/signals#use-cases-for-effects

frontend/ui/src/app/deployments/deployment-status-modal/deployment-status-modal.component.html
```diff
@@ -0,0 +1,87 @@
+<div class="relative w-full xl:max-w-256 max-h-full">
+  <!-- Modal content -->
+  <div class="relative bg-white rounded-lg shadow-sm dark:bg-gray-700">
+    <!-- Modal header -->
+    <div class="flex items-center justify-between p-4 md:p-5 border-b border-gray-200 rounded-t dark:border-gray-600">
+      <h3 class="text-lg font-semibold text-gray-900 dark:text-white">
+        Status History: {{ deploymentTarget().name }}
```
christophenne: maybe now "Deployment Details" or something like that? instead of "Status History"

frontend/ui/src/app/deployments/deployment-status-modal/deployment-status-modal.component.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public readonly deploymentTarget = input.required<DeploymentTarget>();
+  public readonly selectedDeployment = input.required<DeploymentWithLatestRevision>();
+  public readonly closed = output<void>();
+
+  protected readonly statuses: Observable<DeploymentStatusTableEntry[]> = toObservable(this.selectedDeployment).pipe(
+    map((d) => d.id),
+    filter((id) => !!id),
+    distinctUntilChanged(),
+    switchMap((id) => this.deploymentStatuses.pollStatuses(id!)),
```
christophenne: doesn't have to be now, but I think it would make sense to write a `filterNonNull` operator that returns only `T` instead of `T | undefined`, then we can avoid the bang operator afterwards. 

something like https://stackoverflow.com/a/62971842

internal/db/deployment_log_record.go
kosmoz: yes I wanted to avoid having to sort in SQL because it was one cause of some performance issues for the status tables in the past

internal/handlers/agent.go
kosmoz: Right now the revision ID is not really used for anything yet but I wanted to put it in place anyways in case we need it in the future. of course the revision ID should only be set to the ID of a revision of the deployment with the deployment ID sent with the same log entry but this is currently unchecked. So, if for example you have deployment A with revision B and deployment C with revision D, it would be possible to save a log entry with deployment A and revision D.

I don't think it's a huge issue right now, but it should be addressed before the revision ID is used for anything I guess. What do you think?

internal/db/deployment_log_record.go
kosmoz: I thought about this too but then we would have to implement some auto scroll-to-bottom which has caused me some pain in previous projects so I decided to keep the "inverted" mode for now. :smile: 

I think it would make sense to make this a follow-up task?

frontend/ui/src/app/deployments/deployment-status-modal/deployment-status-modal.component.ts
kosmoz: yeah good idea

kosmoz: https://linear.app/glasskube/issue/GK-741/create-a-rxjs-filternonnull-operator-function

internal/db/deployment_log_record.go
kosmoz: https://linear.app/glasskube/issue/GK-742/invert-statuslog-entry-list

internal/handlers/agent.go
christophenne: I see, yes I agree

frontend/ui/src/app/deployments/deployment-status-modal/deployment-logs-table.component.ts
kosmoz: I don't think changing this to using `toObservable` is a good solution, because it would have the same problems as using effect basically. 

Found another solution which relies on supplying the list of logs to the getMore function call, so there is no additional state changes necessary

kosmoz:
>     * I didn't get the log collection of kubernetes deployments to work yet, will try again.

Good point, should be fixed now

>     * Maybe a default max count for development would make sense.

I set it to 200

>     * I think it would be nice to have the possibility to enable logs directly when the deployment is created.

https://linear.app/glasskube/issue/GK-743/add-an-option-to-enable-logging-when-creating-a-deployment

>     * The tabs in the modal behave a bit inconsequent. The status history is being polled, the other things not. They could have a reload button maybe? Or just polling too? Or no polling but reload buttons for all the tabs (including status history)?

I know… but… I didn't want to spend even more time on this task… :sweat: Should we create a follow-up task for this or do it now? @pmig?

frontend/ui/src/app/deployments/deployment-status-modal/timeseries-table.component.ts
```diff
@@ -0,0 +1,142 @@
...
+        merge(
+          source.load(),
+          this.showMore$.pipe(
+            map(() => nextBefore),
+            filter((before) => before !== null),
+            switchMap((before) => source.loadBefore(before))
+          )
+        ).pipe(
+          tap((entries) => {
+            console.log({entries, batchSize: source.batchSize, hasMore: entries.length >= source.batchSize});
```
christophenne: debug output should be removed

frontend/ui/src/app/services/deployment-status.service.ts
```diff
@@ -10,8 +11,9 @@ export class DeploymentStatusService {
...
   private readonly httpClient = inject(HttpClient);
 
-  getStatuses(deploymentId: string): Observable<DeploymentRevisionStatus[]> {
-    return this.httpClient.get<DeploymentRevisionStatus[]>(`${this.baseUrl}/${deploymentId}/status`);
+  getStatuses(deploymentId: string, options?: TimeseriesOptions): Observable<DeploymentRevisionStatus[]> {
+    const params = timeseriesOptionsAsParams(options);
+    return this.httpClient.get<DeploymentRevisionStatus[]>(`${this.baseUrl}/${deploymentId}/status`, {params});
   }
 
   pollStatuses(deploymentId: string): Observable<DeploymentRevisionStatus[]> {
```
christophenne: The only usage of `pollStatuses` is an observable that is never subscribed to, so that could be removed and consequently this function too.

frontend/ui/src/app/deployments/deployment-status-modal/deployment-status-table.component.ts
```diff
@@ -1,43 +1,48 @@
...
+  ) {}
+
+  load(): Observable<TimeseriesEntry[]> {
+    return this.svc
+      .getStatuses(this.deploymentId, {limit: this.batchSize})
+      .pipe(map((logs) => logs.map(statusToTimeseriesEntry)));
+  }
+
+  loadAfter(after: Date): Observable<TimeseriesEntry[]> {
+    // TODO: Handle after
```
christophenne: seems to be handled

PR #882, opened 2025-05-07
feat: add support for custom domains and email from address

kosmoz:


internal/handlers/deployment_targets.go
```diff
@@ -181,7 +181,7 @@ func createAccessForDeploymentTarget(w http.ResponseWriter, r *http.Request) {
 		log.Warn("could not update DeploymentTarget", zap.Error(err))
 		sentry.GetHubFromContext(ctx).CaptureException(err)
 		w.WriteHeader(http.StatusInternalServerError)
-	} else if connectUrl, err := buildConnectUrl(deploymentTarget.ID, targetSecret); err != nil {
+	} else if connectUrl, err := buildConnectUrl(deploymentTarget.ID, *auth.CurrentOrg(), targetSecret); err != nil {
```
christophenne, line 184: Not sure why, but I get the following connect url (without the /api path) when using a custom app domain:
```
curl "christoph-gk:8080?targetId=...&targetSecret=..." | docker compose -f - up -d
```

internal/types/organization.go
```diff
@@ -3,25 +3,54 @@ package types
...
+	RegistryDomain   *string   `db:"registry_domain" json:"registryDomain"`
+	EmailFromAddress *string   `db:"email_from_address" json:"emailFromAddress"`
 }
 
 func (org *Organization) HasFeature(feature Feature) bool {
 	return slices.Contains(org.Features, feature)
 }
 
+func (o Organization) AppDomainOrDefault() string {
+	if o.AppDomain != nil {
```
christophenne, line 34: If a custom domain is set, the function returns the value without the protocol part, whereas it is included in the `env.Host()` default. I think it should be with the protocol in both cases (or at least consistent for both cases)

internal/handlers/agent.go
```diff
@@ -61,6 +61,7 @@ func connectHandler() http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		log := internalctx.GetLogger(ctx)
+		auth := auth.Authentication.Require(ctx)
```
christophenne: This panics because the connect endpoint is not authenticated like the others. We only have the deployment target at this point and from that the organization has to fetched.

internal/types/organization.go
kosmoz: I thought it would be OK to put the scheme in the DB column as well but I understand that this could cause some confusion and errors so I added some logic to coerce the returned value to always have a scheme

internal/handlers/deployment_targets.go
kosmoz: possibly caused by the issue below? please check again

christophenne: yes now it works

PR #899, opened 2025-05-13
feat: add host system metrics collection for deployment targets

christophenne:


cmd/agent/docker/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+	reportFunc func(event *componentstatus.Event)
+}
+
+func (nh *noopHost) GetExtensions() map[component.ID]component.Component {
+	return nil
+}
+
+func (nh *noopHost) Report(event *componentstatus.Event) {
+	nh.reportFunc(event)
+}
```
kosmoz: I don't understand this… 
1. I wouldn't call it "noop" if it actually does something and
2. I don't understand where this is even called

I think a more idiomatic way to implement the Host interface like this would be something like this:
```go
type HostFunc func(event componentstatus.Event)

func (hf HostFunc) GetExtensions() map[component.ID]component.Component {
  return nil
}

func (hf HostFunc) Report(event *componentstatus.Event) { hf(event) }

```

api/agent.go
```diff
@@ -43,3 +44,10 @@ type AgentDeploymentStatus struct {
 	Type       types.DeploymentStatusType `json:"type"`
 	Message    string                     `json:"message"`
 }
+
+type AgentDeploymentTargetMetrics struct {
+	CPUCoresM   int64   `json:"cpuCoresM" db:"cpu_cores_m"`
```
kosmoz: Just out of curiosity, why is it called "coresM"? What does "M" stand for?

EDIT: I just saw that it is multiplied by 1000 below. Maybe rename it to "millis"? Also, why does it need to be sent with this multiplier anyways?

cmd/agent/docker/metrics.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil
+	})
+	if err != nil {
+		logger.Error("failed to create metrics consumer", zap.Error(err))
+	}
+
+	metrics, err = factory.CreateMetrics(ctx, receiver.Settings{
+		ID: component.NewID(factory.Type()),
+		TelemetrySettings: component.TelemetrySettings{
+			MeterProvider:  otel.GetMeterProvider(), // TODO no idea check again
```
kosmoz: Is this TODO still relevant? If yes please add some more info for the rest of us :sweat_smile:

cmd/agent/kubernetes/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var metricsLastReported *time.Time
+
+func reportMetricsIfOutdated(ctx context.Context) {
+	if metricsLastReported != nil && (*metricsLastReported).After(time.Now().Add(-30*time.Second)) {
+		return
+	}
+	go doReportMetrics(ctx)
```
kosmoz: I have concerns here what happens if the doReportMetrics function takes too long. It might cause many metrics collection goroutines to run at the same time, especially if a cluster has many nodes. 

Would it be possible to spawn a goroutine with it's own ticker and loop instead?
```go
func main() {
  go WatchMetrics(ctx)
}

func WatchMetrics(ctx context.Context) {
  tick := time.Tick(30*time.Second)
  for {
    select {
      case <-ctx.Done():
        return
      case <-tick:
        doReportMetrics(ctx)
    }
  }
}
```

frontend/ui/src/app/services/deployment-target-metrics.service.ts
```diff
@@ -0,0 +1,39 @@
...
+  id: string;
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DeploymentTargetsMetricsService {
+  private readonly deploymentTargetMetricsBaseUrl = '/api/v1/deployment-target-metrics';
+  private readonly httpClient = inject(HttpClient);
+  private readonly cache = new DefaultReactiveList<DeploymentTargetLatestMetrics>(
```
kosmoz: Is the cache even used here?

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.ts
```diff
@@ -57,6 +72,9 @@ import {DeploymentModalComponent} from '../deployment-modal.component';
     ConnectInstructionsComponent,
     ReactiveFormsModule,
     DeploymentModalComponent,
+    BytesPipe,
```
kosmoz: unused I think

internal/agentmanifest/common.go
```diff
@@ -66,6 +68,7 @@ func getTemplateData(
 		"manifestEndpoint":  manifestEndpoint,
 		"resourcesEndpoint": resourcesEndpoint,
 		"statusEndpoint":    statusEndpoint,
+		"metricsEndpoint":   metricsEndpoint,
```
kosmoz, line 71: still has to be added in the kubernetes manifest template

kosmoz (commented):
Looks good, but I noticed something strange with the CPU metrics. The value shown in the UI consisently seems higher than the one logged by the docker agent. For example:
image
vs
```

2025-05-14T11:46:09.048+0200    DEBUG   docker/metrics.go:117   cpu usage       {"usage": 0.03876069712526671, "cores": 20}
2025-05-14T11:46:09.048+0200    DEBUG   docker/metrics.go:119   memory usage    {"usage": 0.26662258320931087, "total": 24857772032}
```

I suspect that the value shown in the UI still needs to be divided by the number of cores maybe? In any case the number logged by the agent seems more in line with what other reporting applications show.

The kubernetes agent tends to under-report the metrics probably because the metrics API only reports resource usage by pods, as opposed to the entire system (I guess?)… But for now at least I think this is still fine

api/agent.go
christophenne: I will rename it. 

Because in kubernetes it could be `0.1` CPUs, and instead of sending a float number, I thought it's a bit more elegant/pragmatic to just sent it in the smallest possible unit, like we also do with bytes.

frontend/ui/src/app/services/deployment-target-metrics.service.ts
christophenne: Good point, no.

cmd/agent/docker/metrics.go
christophenne: I just saw that the Report function isn't even part of the `Host` interface. I removed it and now it should be an actual `noop` host I think (and I renamed it to `defaultHost`).

cmd/agent/docker/metrics.go
christophenne: Not really, it's just that the whole telemetry setup seems mysterious, but it also doesn't work without these providers. And according to the docs, some noop provider is being used as a fallback.

frontend/ui/src/styles/charts.scss
kosmoz, line 1: Would it be possible to move these styles to the component where they are used?

kosmoz (commented):
Just one more think, apart from that looks good to me

kosmoz:
also, please run `go mod tidy`, it has some changes

kosmoz (dismissed):

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-metrics.component.html
```diff
@@ -1,13 +1,11 @@
...
 <div class="flex space-x-1 items-center justify-end rtl:space-x-reverse">
   <div class="flex flex-col justify-center items-center">
     <div class="relative">
-      <div class="gauge z-0 percent-{{ cpu }} relative">
+      <div class="gauge z-0 percent-{{ getPercentClass(metrics().cpuUsage) }} relative">
         <div class="gauge-center bg-white dark:bg-gray-800"></div>
       </div>
       <div class="absolute inset-0 flex items-center justify-center text-xs font-thin text-black dark:text-white">
-        {{ cpu }}%
+        {{ getPercent(metrics().cpuUsage) }}%
```
kosmoz: OK sorry for another "one more thing" :see_no_evil: but I think at this point it would make sense to use angulars built-in `percent` pipe for this: https://angular.dev/api/common/PercentPipe

kosmoz (approved):

PR #901, opened 2025-05-14
fix(ui): optimize deployments view for smaller devices

pmig:
I also introduced new 3xl and 4xl breakpoints for full width pages like the dashboard

### DIFF
--- a/frontend/ui/src/styles/theme.scss
+++ b/frontend/ui/src/styles/theme.scss
@@ -22,6 +22,9 @@
   --font-sans: 'Inter', 'system-ui', 'sans-serif';
   --font-display: 'Poppins', 'Inter', 'system-ui', 'sans-serif';
   --font-mono: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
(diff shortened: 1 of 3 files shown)


frontend/ui/src/styles/theme.scss
```diff
@@ -22,6 +22,9 @@
   --font-sans: 'Inter', 'system-ui', 'sans-serif';
   --font-display: 'Poppins', 'Inter', 'system-ui', 'sans-serif';
   --font-mono: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier New', monospace;
+
+  --breakpoint-3xl: 120rem;
```
christophenne: The difference between the `large` breakpoints seems to always be `16rem`, see https://tailwindcss.com/docs/responsive-design#overview – so it would make sense if these new breakpoints where `112rem` and `128rem`? How did you come up with yours? Because they feel a bit better for the dashboard, but maybe that should be changed on the dashboard then.

pmig: yes

PR #924, opened 2025-05-19
feat(ui): add bulk actions for application versions

kosmoz:


internal/handlers/applications.go
```diff
@@ -110,6 +113,73 @@ func updateApplication(w http.ResponseWriter, r *http.Request) {
...
+		log := internalctx.GetLogger(ctx)
+		auth := auth.Authentication.Require(ctx)
+		existing := internalctx.GetApplication(ctx)
+		patch, err := JsonBody[api.PatchApplicationRequest](w, r)
+		if err != nil {
+			return
+		}
+
+		if err := db.RunTx(ctx, func(ctx context.Context) error {
+			appliationNeedsUpdate := false
```
christophenne, line 128: any specific reason to use these two `...NeedsUpdate` flags instead of just calling the `db.Update...` function directly in these ifs like 

```
if patch.Name != nil && patch.Name != &existing.Name {
    existing.Name = *patch.Name
    if err := db.UpdateApplication(...
}
```

internal/handlers/applications.go
```diff
@@ -110,6 +113,73 @@ func updateApplication(w http.ResponseWriter, r *http.Request) {
...
+			}
+			return nil
+		}); errors.Is(err, pgx.ErrTxCommitRollback) {
+			log.Warn("could not commit db transaction", zap.Error(err))
+			sentry.GetHubFromContext(ctx).CaptureException(err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		RespondJSON(w, existing)
```
christophenne, line 181: only if err == nil I think?

frontend/ui/src/app/applications/application-detail.component.ts
```diff
@@ -326,6 +332,35 @@ export class ApplicationDetailComponent implements OnInit, OnDestroy {
       .subscribe();
   }
 
+  async bulkArchiveVersions(application: Application) {
```
christophenne:
```suggestion
  async bulkArchiveVersions() {
```

christophenne: or alternatively use the parameter instead of `switchMap(() => this.application$)` a few lines below

internal/handlers/applications.go
kosmoz: I think this makes adding more "patchable" properties easier… no other reason

frontend/ui/src/app/applications/application-detail.component.ts
kosmoz: yes of course lol

PR #927, opened 2025-05-20
fix(ui): deployments page handles archived version scenarios better

christophenne:


frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -9,6 +9,7 @@ import {
   catchError,
   combineLatest,
   combineLatestWith,
+  filter,
```
kosmoz:
```suggestion
```

frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -114,14 +115,23 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
...
+              )
             )
           )
           .flat()
       )
     ),
+    map((dts) => dts.filter((dt) => !!dt)),
     map((result) =>
-      result.filter((it) => it.version?.id !== it.d.applicationVersionId).map((it) => ({...it, version: it.version!}))
+      result.filter((it) => it.version.id !== it.d.applicationVersionId).map((it) => ({...it, version: it.version}))
```
kosmoz: I think this is no longer necessary
```suggestion
      result.filter((it) => it.version.id !== it.d.applicationVersionId))
```

frontend/ui/src/app/deployments/deployment-targets.component.ts
```diff
@@ -114,14 +115,23 @@ export class DeploymentTargetsComponent implements AfterViewInit, OnDestroy {
...
+                  }
+                  return undefined;
+                })
+              )
             )
           )
           .flat()
       )
     ),
+    map((dts) => dts.filter((dt) => !!dt)),
```
kosmoz: I've kind of started preferring this more explicit style recently, especially when handling basic types because stuff like `!!'' === false` (which is not the case here but still)… what do you think?

```suggestion
    map((dts) => dts.filter((dt) => dt !== undefined)),
```

christophenne: yes good idea

kosmoz (approved):

PR #941, opened 2025-05-22
fix(backend): lazily initialize jwt auth

christophenne:
@kosmoz not sure how clean this is for the references in `internal/auth/authentication.go`, since these are also just vars, and they could be used before `env.Initialize()`. I think they are not because these are just used in handlers, and handlers are only called once the whole setup has finished, but its still a bit of a pitfall.

### DIFF
--- a/internal/authjwt/jwt.go
+++ b/internal/authjwt/jwt.go
@@ -36,7 +36,7 @@ const (
 // which should be OK for now.
 //
 // TODO: Maybe migrate to asymmetric encryption at some point.
-var JWTAuth = sync.OnceValue[*jwtauth.JWTAuth](func() *jwtauth.JWTAuth {
+var JWTAuth = sync.OnceValue(func() *jwtauth.JWTAuth {
 	return jwtauth.New("HS256", env.JWTSecret(), nil)
 })
 


internal/authjwt/jwt.go
```diff
@@ -35,7 +36,9 @@ const (
 // which should be OK for now.
 //
 // TODO: Maybe migrate to asymmetric encryption at some point.
-var JWTAuth = jwtauth.New("HS256", env.JWTSecret(), nil)
+var JWTAuth = sync.OnceValue[*jwtauth.JWTAuth](func() *jwtauth.JWTAuth {
```
kosmoz:
```suggestion
var JWTAuth = sync.OnceValue(func() *jwtauth.JWTAuth {
```

kosmoz:
> @kosmoz not sure how clean this is for the references in `internal/auth/authentication.go`, since these are also just vars, and they could be used before `env.Initialize()`. I think they are not because these are just used in handlers, and handlers are only called once the whole setup has finished, but its still a bit of a pitfall.

True, but I think it should be fine and it's definitely good enough for now

kosmoz (approved):

PR #948, opened 2025-05-23
fix(agent): fix kubernetes agent sending progressing updates during status check

kosmoz:


cmd/agent/docker/main.go
```diff
@@ -31,6 +32,11 @@ func init() {
 func main() {
 	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
 
+	logger.Info("kubernetes agent is starting",
```
christophenne:
```suggestion
	logger.Info("docker agent is starting",
```

PR #959, opened 2025-05-26
feat: users can be part of multiple organizations

christophenne:


frontend/ui/src/app/components/users/users.component.ts
```diff
@@ -153,17 +159,15 @@ export class UsersComponent implements OnDestroy {
...
-          if (e instanceof HttpErrorResponse && e.status === 400) {
-            this.toast.error(
-              `User ${user.name ?? user.email} cannot be deleted.
-              Please ensure there are no deployments managed by this user and try again.`
-            );
-          }
+          const msg = getFormDisplayedError(e);
+          if (msg) {
+            this.toast.error(msg);
+          } // ??
```
kosmoz: please remove or explain :-)
```suggestion
          }
```

internal/handlers/auth.go
```diff
@@ -165,29 +214,29 @@ func authResetPasswordHandler(w http.ResponseWriter, r *http.Request) {
...
 		mailOpts := []mail.MailOpt{
 			mail.To(user.Email),
 			mail.Subject("Password reset"),
-			mail.HtmlBodyTemplate(mailtemplates.PasswordReset(*user, org, token)),
 		}
-		if from, err := customdomains.EmailFromAddressParsedOrDefault(org); err == nil {
-			mailOpts = append(mailOpts, mail.From(*from))
-		} else {
-			log.Warn("error parsing custom from address", zap.Error(err))
+		if len(orgs) > 0 {
```
kosmoz, line 231: should `len(orgs) == 0` be an error here? I think otherwise it might cause issues further down

internal/migrations/sql/42_tutorials_organization_scoped.up.sql
```diff
@@ -0,0 +1,16 @@
+-- this migration assumes that every user is part of exactly one organization
+-- if there is a user without an organization, this will fail. a manual fix would be to delete the corresponding UserAccount_TutorialProgress
+
+ALTER TABLE UserAccount_TutorialProgress ADD COLUMN organization_id UUID;
```
kosmoz: should this have a foreign key constraint?

internal/migrations/sql/41_pat_organization_scoped.up.sql
```diff
@@ -0,0 +1,13 @@
+-- this migration assumes that every user is part of exactly one organization
+-- if there is a user without an organization, this will fail. a manual fix would be to delete the corresponding AccessToken
+
+ALTER TABLE AccessToken ADD COLUMN organization_id UUID;
```
kosmoz: should this have a foreign key constraint?

kosmoz (commented):
Looks good, I think there are some things we might want to improve over time but this looks like a strong start!

Some minor findings during review…

internal/handlers/auth.go
christophenne: The idea was that this can happen and the user can still request a pw reset. Further down the `org = nil` case is handled inside `mailtemplates.PasswordReset`.

frontend/ui/src/app/components/nav-bar/nav-bar.component.html
```diff
@@ -5,101 +5,231 @@
...
+                            [title]="'Switch to this organization'"
+                            class="flex items-center w-full text-start px-4 py-2 text-sm text-gray-700 hover:bg-gray-100 dark:text-gray-300 dark:hover:bg-gray-600 dark:hover:text-white"
+                            role="menuitem">
+                            <fa-icon class="me-4" [icon]="faShuffle"></fa-icon>
+                            <span
+                              class="flex flex-col text-start text-sm font-medium text-gray-900 dark:text-gray-300"
+                              role="none">
+                              {{ org.name }}
+                              <span class="text-xs font-light text-gray-500 dark:text-gray-400" role="none">
+                                {{ org.userRole | titlecase }}
```
pmig: I would prefer to only add the role "vendor" if the user has the role vendor, if the user has the role customer, I would not write any role, as for the end customer that might be a litte confusing if he is in multiple orgs "customer"

pmig (changes requested):
It's such a great improvement!

I could only find one bug regarding the images, where the images couldn't get loaded.

1. Create a new org
2. Invite an existing user (that has an image) to the new org as
3. In the dashboard Distr tries to load the profile image, but fails

frontend/ui/src/app/components/nav-bar/nav-bar.component.html
christophenne: Fair, for only-customers it makes no sense. Changed it to show the role only if the user is vendor in at least one organization.

pmig (approved):

internal/handlers/auth.go
kosmoz: Ah ok all good then :+1:

kosmoz (commented):
Alright looks good to me!

The only thing we might want to still improve before merging is the visuals of the organization picker dropdown for the case where there is no organization to switch to, as it will be the most common case probably (at least in the beginning) and it looks a little unpolished:

image

My suggestion would be to either:
* also show the current org in the list, so there is always at least one
* hide the top part and horizontal line if there are no orgs to switch to
* add a muted text like "You don't have any other organizations" or something like that above the horizontal line

pmig:
> The only thing we might want to still improve before merging is the visuals of the organization picker dropdown for the case where there is no organization to switch to, as it will be the most common case probably (at least in the beginning) and it looks a little unpolished:

I would vote for option 1 or 3  having the "self" organization in the list might be confusing if there are more orgs.

For option 3 I would go with:

> You are not part of any other organization. You can create a additional organization below.

kosmoz:
> For option 3 I would go with:
> 
> > You are not part of any other organization. You can create an additional organization below.

IMO only the first sentence would be better, as space is quite limited in the dropdown and the affordance of the "+ Create Organization" button is quite obvious

kosmoz (approved):

PR #969, opened 2025-05-27
ci: group oTel dependencies

pmig:


renovate.json
```diff
@@ -50,6 +50,19 @@
       "schedule": [
         "* 6-9 * * 1"
       ]
+    },
+    {
+      "extends": [
+        "monorepo:opentelemetry-go",
+        "monorepo:opentelemetry-go-contrib"
```
kosmoz: Does this mean that these two will be grouped together?

kosmoz: because I think it would be better to keep them separate

pmig: yes, but happy to have two groups. I'll improve

PR #974, opened 2025-05-28
chore(deploy): add cleanup jobs config in deploy/docker and helm chart

kosmoz:


deploy/docker/.env
```diff
@@ -60,3 +60,9 @@ USER_EMAIL_VERIFICATION_REQUIRED=false
 # SENTRY_DEBUG=true
 # FRONTEND_SENTRY_DSN="..."
 # FRONTEND_SENTRY_TRACE_SAMPLE_RATE=1.0
+
+# Scheduled job config
+CLEANUP_DEPLOYMENT_REVISION_STATUS_CRON="*/5 * * * *"
+CLEANUP_DEPLOYMENT_TARGET_STATUS_CRON="*/5 * * * *"
+CLEANUP_DEPLOYMENT_TARGET_METRICS_CRON="*/5 * * * *"
+CLEANUP_DEPLOYMENT_LOG_RECORD_CRON="*/5 * * * *"
```
christophenne:
```suggestion

LOG_RECORD_ENTRIES_MAX_COUNT=500

# Scheduled job config
# cron interval in which revision statuses older than STATUS_ENTRIES_MAX_AGE will be deleted
CLEANUP_DEPLOYMENT_REVISION_STATUS_CRON="*/5 * * * *" 
# cron interval in which deployment target statuses older than STATUS_ENTRIES_MAX_AGE will be deleted
CLEANUP_DEPLOYMENT_TARGET_STATUS_CRON="*/5 * * * *" 
# cron interval in which metrics older than METRICS_ENTRIES_MAX_AGE will be deleted
CLEANUP_DEPLOYMENT_TARGET_METRICS_CRON="*/5 * * * *" 
# cron interval in which log entries older than the last LOG_RECORD_ENTRIES_MAX_COUNT will be deleted
CLEANUP_DEPLOYMENT_LOG_RECORD_CRON="*/5 * * * *" 
```

christophenne: just a suggestion to do it here, but I think it should be stated somewhere that these variables are related

kosmoz:
> I think the chart README should also be regenerated at some time :)

Good idea, but I have no clue what tool was originally used to generate it… @pmig you remember by any chance?

deploy/docker/.env
kosmoz: good point. maybe something for the self-hosting docs?

kosmoz:
> > > I think the chart README should also be regenerated at some time :)
> > 
> > 
> > Good idea, but I have no clue what tool was originally used to generate it… @pmig you remember by any chance?
> 
> According to the git history it was
> 
> ```
> Autogenerated from chart metadata using [helm-docs v1.14.2](https://github.com/norwoodj/helm-docs/releases/v1.14.2)
> ```
> 
> but my commit back then seems to have removed this, and I don't remember if I used the same one (and why it removed that line) or another 😅

OK thanks, I updated the readme and added helm-docs to the tools.toml file

PR #1016, opened 2025-06-05
feat(mcp): add initial version of the Distr MCP server

kosmoz:


cmd/mcp/main.go
```diff
@@ -0,0 +1,107 @@
...
+	sse bool
+}{}
+
+func serveCmdRun(cmd *cobra.Command, args []string) {
+	clientConfig, err := clientConfigFromEnv()
+	if err != nil {
+		log.Fatal("client config is invalid", zap.Error(err))
+	}
+
+	log.Sugar().Debugf("got client config: %v", clientConfig)
```
christophenne, line 60: not sure if the token should be printed here

cmd/mcp/client/deployment_targets.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	_, err = httpstatus.CheckStatus(c.config.httpClient.Do(req))
+	return err
+}
+
+func (c *DeploymentTargets) Connect(
+	ctx context.Context,
+	id uuid.UUID,
+) (*api.DeploymentTargetAccessTokenResponse, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.url(id.String()), nil)
```
christophenne:
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.url(id.String(), "access-request"), nil)
```

I think

cmd/mcp/main.go
kosmoz: This would only be printed in dev mode (buildconfig.IsDevelopment() == true), because otherwise the MCP uses a production logger which doesn't do Debug. So I thought it would be fine… during development, it was pretty useful to see the token in the logs. but maybe we can also partly redact the token during stringification? :thinking:

christophenne: Oh I didn't catch that. Then it's fine I think.

kosmoz: Since I think you raised a valid point and it's probably a good idea to try to prevent accidental credential logging I added the aforementioned partial redaction for API keys

PR #1039, opened 2025-06-11
feat: support login with github, google, microsoft

christophenne:


internal/handlers/auth_oidc.go
```diff
@@ -0,0 +1,324 @@
...
+	"golang.org/x/oauth2/google"
+	"golang.org/x/oauth2/microsoft"
+)
+
+const redirectToLoginOIDCFailed = "/login?reason=oidc-failed"
+
+func AuthOIDCRouter(r chi.Router) {
+	ctx := context.Background()
+	if env.OIDCGoogleEnabled() {
+		// TODO proper (lazy?) initialization (every server instance makes the requests)
```
christophenne: not sure what is the best way to deal with this setup and where to do it? and if it should even panic in case of an error...

kosmoz: Would it make sense to do it in `svc`? Then the error could be propagated from there. 

Why does it need to be lazy?

kosmoz: Would it make sense to do it in `svc`? Then the error could be propagated from there. 

Why does it need to be lazy?

kosmoz: Maybe it would be possible to assemble a `map[string]OAuthProvider` where `OAuthProvider` is an abstraction with multiple possible implementations: `IDVerifierProvider`, `GitHubAccessTokenProvider`, something like that. 

Would that make sense?

christophenne: I did it via `svc`, but I am not sure.

internal/oidc/oidc.go
```diff
@@ -0,0 +1,181 @@
...
+		if err != nil {
+			return "", false, fmt.Errorf("failed to get email from github: %w", err)
+		}
+		return email, true, nil
+	}
+	return "", false, fmt.Errorf("unsupported OIDC provider: %s", provider)
+}
+
+func getEmailFromGithubAccessToken(accessToken string) (string, error) {
+	req, err := http.NewRequest("GET", "https://api.github.com/user/emails", nil)
```
kosmoz: Better use NewRequestWithContext instead. Also maybe use the `http.MethodGet` constant

internal/oidc/oidc.go
```diff
@@ -0,0 +1,181 @@
...
+	prov := o.providers[provider]
+	if prov == nil || prov.OAuth2Config == nil {
+		return "", false, fmt.Errorf("OIDC provider not configured: %s", provider)
+	}
+	token, err := prov.OAuth2Config(r).Exchange(ctx, code)
+	if err != nil {
+		return "", false, fmt.Errorf("token exchange failed: %w", err)
+	}
+
+	if prov.IDTokenVerifier != nil {
```
kosmoz: This distinction feels kind of redundant here, since we already know what provider is being used. it should be possible to declare which provider uses which email "extraction" method at when configuring the `providerContext`s. 

Maybe by specifying a "emailExtractorFunc" (`func(ctx, token) (string, error)`), similar to the `OAuth2Config` property. Then the providerContext doesn't require knowledge of the IDVerifier, it can be encapsulated with a higher order function like `func idVerifierEmailExtractorFunc(verifier) func(ctx, token) (string, error) { ... }`.

Then the Microsoft (and Google) provider context can be created like this:
```go
microsoftVerifier := microsoftProvider.Verifier(config)
providerContext{
	OAuth2Config:    getMicrosoftOauth2Config,
	IDTokenVerifier: idVerifierEmailExtractorFunc(microsoftVerifier),
}
```

And the GitHub provider like this:
```go
providerContext{
	OAuth2Config:    getGithubOauth2Config,
	IDTokenVerifier: getEmailFromGithubAPI,
}
```

What do you think?

Sorry if this seems a little nitpicky already, but I think getting this part right now will hopefully make it easier to add more providers in the future…

internal/oidc/oidc.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func getEmailFromGithubAccessToken(accessToken string) (string, error) {
+	req, err := http.NewRequest("GET", "https://api.github.com/user/emails", nil)
+	if err != nil {
+		return "", err
+	}
+	req.Header.Set("Authorization", "Bearer "+accessToken)
+
+	client := &http.Client{}
```
kosmoz: any reason not to use `http.DefaultClient`?

internal/types/types.go
```diff
@@ -22,6 +22,7 @@ type (
 	DockerType            string
 	Tutorial              string
 	FileScope             string
+	OIDCProvider          string
```
kosmoz: I'm not sure if this type is necessary. Personally I think I would remove it and just use string constants. 

Regardless though, I think the constants (and this type if you prefer to keep it) should be moved to the `oidc` module

internal/oidc/oidc.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&emails); err != nil {
+		return "", err
+	}
+
+	for _, email := range emails {
+		if email.Primary && email.Verified {
+			return email.Email, nil
+		}
+	}
```
kosmoz: For me personally this does not work, because my work email is not set as my primary email address on GitHub. I don't think there is a perfect solution for this but maybe we could try to find a user for any email address if there is none for the primary one. But then we would also have to deal with the possibility of one GitHub user relating to more than one Distr user …

I understand that this would be a pretty big change so we don't have to fix it now, but I think it should definitely be a follow up task because I suspect this case could be quite common

kosmoz (commented):
Here are some thoughts. Generally, the feature worked nicely :slightly_smiling_face:

internal/types/types.go
christophenne: I introduced it because using the same string constant too often makes the linter complain, so I will keep it and move it.

internal/oidc/oidc.go
christophenne: Yes, there might have to be a selection step for the user where they choose which one to use. Will create a task.

internal/oidc/oidc.go
christophenne: didn't know it existed :)

internal/oidc/oidc.go
christophenne: Yes, good idea.

internal/types/types.go
kosmoz: Oh yes, the constants are definitely not a bad idea, I was just suggesting removing the custom type. but either way is fine for me

kosmoz (dismissed):

internal/env/env.go
```diff
@@ -190,6 +191,12 @@ func Initialize() {
 		oidcMicrosoftClientSecret = util.PtrTo(envutil.RequireEnv("OIDC_MICROSOFT_CLIENT_SECRET"))
 		oidcMicrosoftTenantID = util.PtrTo(envutil.RequireEnv("OIDC_MICROSOFT_TENANT_ID"))
 	}
+	wellKnownMicrosoftIdentityAssociationEnc := envutil.GetEnvOrNil("WELLKNOWN_MICROSOFT_IDENTITY_ASSOCIATION_JSON_BASE64")
```
kosmoz: I think it would be easier to store `wellKnownMicrosoftIdentityAssociation` as `[]byte` and use `envutil.GetEnvParsedOrNil`

kosmoz: or rather `GetEnvParsedOrDefault` … see e.g. `agentDockerConfig`

kosmoz (approved):

kosmoz (dismissed):

pmig (approved):

frontend/ui/src/app/login/login.component.html
```diff
@@ -1,72 +1,112 @@
...
+                  Or use one of these to sign in:
+                </p>
+                <div class="flex flex-row justify-center w-full gap-2">
+                  @if (loginConfig.oidcGithubEnabled) {
+                    <a
+                      [href]="getLoginURL('github')"
+                      type="button"
+                      class="w-full md:w-auto flex items-center justify-center py-2 px-4 text-sm font-medium text-gray-900 focus:outline-none bg-white rounded-lg border border-gray-200 hover:bg-gray-100 hover:text-primary-700 focus:z-10 focus:ring-4 focus:ring-gray-200 dark:focus:ring-gray-700 dark:bg-gray-800 dark:text-white dark:border-gray-600 dark:hover:text-white dark:hover:bg-gray-700">
+                      <fa-icon class="w-4 h-4 me-1 flex align-middle" [icon]="faGithub"> </fa-icon>
+                      Github
```
pmig:
```suggestion
                      GitHub
```

pmig (changes requested):

kosmoz (approved):

pmig (approved):

PR #1158, opened 2025-09-10
feat(registry): add GCS compatibility for registry storage

Demonsthere:
This PR is based on https://stackoverflow.com/a/77189361 and allows Distr to use GCS (Google Buckets) as a S3 backend. GCS is compatible with the S3 protocol already, however some changes to the v2 of the sdk require tinkering in order to make it work.
This PR:
- chore: updates the aws sdk to the latests version
- fix: add missing postgres config to allow the db to listen to connections
- feat: add a `ResignForGCP` Middleware repackaging the requests to make them GCS compliant

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4990,9 +4990,9 @@
       "optional": true
     },
     "node_modules/@types/jasmine": {
-      "version": "5.1.8",
-      "resolved": "https://registry.npmjs.org/@types/jasmine/-/jasmine-5.1.8.tgz",
-      "integrity": "sha512-u7/CnvRdh6AaaIzYjCgUuVbREFgulhX05Qtf6ZtW+aOcjCKKVvKgpkPYJBFTZSHtFBYimzU4zP0V2vrEsq9Wcg==",
+      "version": "5.1.9",
+      "resolved": "https://registry.npmjs.org/@types/jasmine/-/jasmine-5.1.9.tgz",
+      "integrity": "sha512-8t4HtkW4wxiPVedMpeZ63n3vlWxEIquo/zc1Tm8ElU+SqVV7+D3Na2PWaJUp179AzTragMWVwkMv7mvty0NfyQ==",
       "dev": true,
       "license": "MIT"
     },


deploy/charts/distr/values.yaml
```diff
@@ -239,6 +246,12 @@ postgresql:
     database: distr
     existingSecret: ''
   architecture: standalone
+  primary:
```
kosmoz: I'm not sure this is a good idea to keep around. Is there a specific reason why you added this?

internal/registry/blob/s3/handler.go
```diff
@@ -43,13 +43,18 @@ var (
...
 		s3Client = s3.New(s3.Options{}, clientOpts(s3Config))
+		if s3Config.ResignForGCP {
+			s3Client = s3.New(s3.Options{}, clientOpts(s3Config), ResignForGCP)
+		}
 	} else {
-		otelaws.AppendMiddlewares(&awsconfig.APIOptions)
-		s3Client = s3.NewFromConfig(awsconfig, clientOpts(s3Config))
+		otelaws.AppendMiddlewares(&config.APIOptions)
+		s3Client = s3.NewFromConfig(config, clientOpts(s3Config))
+		if s3Config.ResignForGCP {
```
kosmoz, line 54: In my understanding, `err == nil` here implies that a valid AWS config was found. In that case, is there a valid use-case where it makes sense to add the GCS specific behavior?

kosmoz (commented):
Hi @Demonsthere, thanks for this! Looks good for the most part, but I have a few questions (see below).

Also, it looks like the build failed due to some lint errors. Can you please check those?

deploy/charts/distr/values.yaml
Demonsthere: Without this it is not possible to use the db, even for a demo, as only connections from localhost are permitted. We need to allow distr to talk to the db. This allows that by default making the experience of installing distr with postgres much smoother. If you don't want to set it by default, then i would encourage you to add that in the docs, as currently just installing the chart with the following config simply does not work
```yaml
postgresql:
  enabled: true
```

internal/registry/blob/s3/handler.go
Demonsthere: I tried to follow the existing patten, but to be honest it is a little confusing. I would expect that if err != nil we would throw and error here/panic as the client cannot be initialised. However instead `s3Client = s3.New(s3.Options{}, clientOpts(s3Config))` is run which would overshadow the error from the intial init 🤔

deploy/charts/distr/values.yaml
kosmoz: Thank you for clarifying. It *should* work because the upstream chart writes the database password to a `Secret` resource which we reference in the deployment environment, see here: https://github.com/glasskube/distr/blob/e6636f61aa1810e19056bb2055d680cff4ac3430/deploy/charts/distr/templates/_helpers.tpl#L124-L128

But maybe something has changed in the upstream chart? I will investigate this tomorrow.

internal/registry/blob/s3/handler.go
kosmoz: Let me try to explain: When running on AWS and using regular S3, you might want to use AWS an specific authentication mechanism (e.g. pod identity associations), which is easiest to achieve by using the default config. But we also want to support alternative S3 APIs, which should IMO not use the `AWS_...` environment variables. That's why we try to use the default config and fall back to a client without AWS config if there is an error. 

I would like to add more nuanced error handling, but the AWS SDK doesn't provide errors for better handling AFAIK.

Definitely something that should be documented as comments, thank you for the honest feedback.

Regarding the issue at hand, I think it makes sense to keep the GCS resigning in place since it means more closely adhering to the config specified in environment.

deploy/charts/distr/values.yaml
Demonsthere: For clarification, the password is set correctly, and the chart is using that, however what i found when installing it on a cluster is that the db rejected the connections, as they were coming from an untrusted source. The added configurations allows access from all sources. 
This can be tested by creating a local cluster, installing the chart and verifying if the hub can speak to the db. Dunno if you have such a test in the CI, or are interested in adding it :)

kosmoz: Thanks, I just tried installing the chart from `main` in a minikube cluster and it does seem to work for me:
```shell
minikube start
minikube addons enable ingress
helm upgrade --install distr . --set postgresql.enabled=true --set minio.enabled=true --set ingress.enabled=true
```

The pod restarts a couple of times initially until the DB is ready but it eventually becomes ready too. 

Did you perhaps use a different Kubernetes distribution (e.g. `kind`)?

Demonsthere: I tested that an a regular GCP kubernetes cluster, and in k3d afair 🤔 Will try to reproduce it, but maybe it was environment specific

Demonsthere: Ok, seems that i was mistaken here, maybe those were some temporary errors related to the startup? Out of the box looks like it is working now 
```shell
postgresql 10:34:18.70 INFO  ==> ** Starting PostgreSQL **
2025-09-11 10:34:18.827 GMT [1] LOG:  pgaudit extension initialized
2025-09-11 10:34:18.913 GMT [1] LOG:  starting PostgreSQL 17.4 on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
2025-09-11 10:34:18.913 GMT [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2025-09-11 10:34:18.913 GMT [1] LOG:  listening on IPv6 address "::", port 5432
2025-09-11 10:34:18.941 GMT [1] LOG:  listening on Unix socket "/tmp/.s.PGSQL.5432"
2025-09-11 10:34:18.952 GMT [154] LOG:  database system was shut down at 2025-09-11 10:34:18 GMT
2025-09-11 10:34:19.016 GMT [1] LOG:  database system is ready to accept connections
2025-09-11 10:34:19.308 GMT [164] LOG:  connection received: host=localhost port=57780
2025-09-11 10:34:29.261 GMT [171] LOG:  connection received: host=localhost port=43172
2025-09-11 10:34:37.112 GMT [172] LOG:  connection received: host=10.213.97.42 port=52174
2025-09-11 10:34:37.115 GMT [173] LOG:  connection received: host=10.213.97.42 port=52186
2025-09-11 10:34:37.119 GMT [173] LOG:  connection authenticated: identity="distr" method=md5 (/opt/bitnami/postgresql/conf/pg_hba.conf:1)
2025-09-11 10:34:37.119 GMT [173] LOG:  connection authorized: user=distr database=distr
2025-09-11 10:34:38.818 GMT [174] LOG:  connection received: host=10.213.97.42 port=52190
2025-09-11 10:34:38.821 GMT [175] LOG:  connection received: host=10.213.97.42 port=52200
2025-09-11 10:34:38.823 GMT [175] LOG:  connection authenticated: identity="distr" method=md5 (/opt/bitnami/postgresql/conf/pg_hba.conf:1)
2025-09-11 10:34:38.823 GMT [175] LOG:  connection authorized: user=distr database=distr
2025-09-11 10:34:38.918 GMT [173] LOG:  disconnection: session time: 0:00:01.803 user=distr database=distr host=10.213.97.42 port=52186
2025-09-11 10:34:39.314 GMT [183] LOG:  connection received: host=localhost port=34248
2025-09-11 10:34:40.707 GMT [190] LOG:  connection received: host=localhost port=34260
2025-09-11 10:34:49.262 GMT [197] LOG:  connection received: host=localhost port=46074
2025-09-11 10:34:50.708 GMT [204] LOG:  connection received: host=localhost port=46076
2025-09-11 10:34:59.509 GMT [212] LOG:  connection received: host=localhost port=60152
2025-09-11 10:35:00.710 GMT [219] LOG:  connection received: host=localhost port=60158
2025-09-11 10:35:09.263 GMT [226] LOG:  connection received: host=localhost port=49994
2025-09-11 10:35:10.708 GMT [233] LOG:  connection received: host=localhost port=50000
2025-09-11 10:35:19.510 GMT [241] LOG:  connection received: host=localhost port=44430
2025-09-11 10:35:20.711 GMT [248] LOG:  connection received: host=localhost port=44438
2025-09-11 10:35:29.264 GMT [255] LOG:  connection received: host=localhost port=32918
2025-09-11 10:35:30.710 GMT [262] LOG:  connection received: host=localhost port=32920
```

PR #1208, opened 2025-09-15
feat: generic OIDC provider [wip]

Demonsthere:
Hello again!
This is my attempt at adding support for https://github.com/glasskube/distr/discussions/1154.
The goal here is to be able to use dex, as a stand in for any 3rd party OIDC provider, and be able to deploy distr together with dex in docker and helm :)

### DIFF
--- a/deploy/docker/docker-compose.yaml
+++ b/deploy/docker/docker-compose.yaml
@@ -2,6 +2,7 @@ name: distr
 
 services:
   postgres:
+    container_name: postgres
     image: 'postgres:17-alpine'
     environment:
       POSTGRES_USER: ${POSTGRES_USER}
@@ -17,7 +18,9 @@ services:
       retries: 6
       start_period: 30s
   hub:
-    image: ghcr.io/glasskube/distr:1.13.0 # x-release-please-version
+    container_name: distr
+#    image: ghcr.io/glasskube/distr:1.13.0 # x-release-please-version
+    image: ghcr.io/glasskube/distr:snapshot
     depends_on:
       postgres:
         condition: service_healthy
@@ -32,6 +35,7 @@ services:
   # If you don't want to use the OCI registry (REGISTRY_ENABLED=false) or want to use an external S3 provider,
   # you don't need to run the minio service.
   minio:
+    container_name: minio
     image: 'minio/minio:RELEASE.2025-02-28T09-55-16Z'
     entrypoint: sh
     command:
@@ -46,8 +50,17 @@ services:
     volumes:
       - minio:/data/
 
+  # Example SMTP provider
+  mailhog:
+    container_name: mailhog
+    image: mailhog/mailhog:latest
+    ports:
+      - "1025:1025"
+      - "8025:8025"
+
   # OIDC provider mockup
   dex:
+    container_name: dex
     image: dexidp/dex:latest
     ports:
       - "5556:5556"
--- a/deploy/docker/docker-compose.yaml
+++ b/deploy/docker/docker-compose.yaml
@@ -15,6 +15,7 @@ services:
       start_interval: 1s
       interval: 5s
       retries: 6
+      start_period: 30s
   hub:
     image: ghcr.io/glasskube/distr:1.13.0 # x-release-please-version
     depends_on:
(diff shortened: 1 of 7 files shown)


Demonsthere:
So, I think i found the proper place to add the oidc support to the code base and initialize the provider, however I do not see it as an option in the UI 

```
docker logs distr -f                                                        
2025-09-15T13:54:56.469Z	INFO	svc/registry.go:51	initializing service registry	{"version": "snapshot", "commit": "1595aee", "release": false}
2025-09-15T13:54:56.483Z	INFO	migrations/migrate.go:53	migrations completed	{"error": "no change"}
2025-09-15T13:54:56.494Z	INFO	jobs/logger.go:23	gocron: new scheduler created
2025-09-15T13:54:56.494Z	INFO	oidc/oidc.go:96	initializing generic OIDC
2025-09-15T13:54:56.498Z	INFO	jobs/scheduler.go:41	job scheduler starting	{"jobs": 5}
2025-09-15T13:54:56.498Z	DEBUG	jobs/logger.go:13	gocron: scheduler starting
2025-09-15T13:54:56.498Z	INFO	jobs/logger.go:23	gocron: scheduler started
2025-09-15T13:54:56.498Z	INFO	server/main.go:33	starting listener on :8080	{"server": "main"}
2025-09-15T13:54:56.499Z	INFO	server/main.go:33	starting listener on :8585	{"server": "registry"}
2025-09-15T13:54:56.499Z	DEBUG	jobs/logger.go:13	gocron: executor started
```

image

Any pointers as to where to add it to so the UI will be aware of the new provider type?

Demonsthere:
Ok, found it x)
image

Demonsthere:
E2E example of using https://www.ory.sh as a custom provider.
https://github.com/user-attachments/assets/e44210d0-5b18-477b-ab60-28502ceb506b

As far as I understand, the OIDC login work for Vendors only? Customers still need to be onboarded and invited via email into the system, or can I enable the OIDC for Customers as well?
My expected flow would look like this:
The system has a single Vendor to which users/admins can login via OIDC, but no new Vendor can register (registration is disabled). Customers can be invited or self-service registration for clients is present, and they can login via the same OIDC providers as the vendors. The goal here is to have a one to many distribution pattern, but allow customers to login/register via the oidc provider.

internal/migrations/sql/49_oidc_pkce.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE OIDCState ADD COLUMN pkce_code_verifier VARCHAR(128);
```
kosmoz:
```suggestion
ALTER TABLE OIDCState ADD COLUMN pkce_code_verifier TEXT;
```

We use `TEXT` everywhere else. I don't think there is a security concern in allowing longer values for the challenge verifier so I would prefer to use `TEXT` here too.

internal/handlers/auth_oidc.go
```diff
@@ -87,7 +89,7 @@ func authLoginOidcCallbackHandler(w http.ResponseWriter, r *http.Request) {
 		} else if len(orgs) < 1 {
 			// TODO deduplicate (regular login)
 			org.Name = user.Email
-			org.UserRole = types.UserRoleVendor
+			org.UserRole = env.DefaultUserRole()
```
kosmoz: I struggle to imagine a situation where it would make sense to change the default role to `customer`. The user would end up in an empty organization where they can't really do anything.

Is there a reason this was added that I'm missing?

Demonsthere:
> I don't think dex and mailhog should be present in both the helm chart and default compose file. However, maybe we could create an examples directory with a working compose file? In any case I would consider this should be done in a separate PR, could you please revert those changes for now?

No worries, i added them mostly for easier local development :) The pr needs a cleanup anyway!

internal/handlers/auth_oidc.go
Demonsthere: No, this is a leftover from something i was testing, and can be discarded

deploy/docker/docker-compose.yaml
```diff
@@ -2,6 +2,7 @@ name: distr
 
 services:
   postgres:
+    container_name: postgres
```
kosmoz: One more thing that slipped through the previous review, could you also please revert the container_name changes? I don't think they are necessary. start_period is fine we can keep that.

kosmoz (changes requested):

deploy/docker/docker-compose.yaml
Demonsthere: yeah that is not required but allows you to get logs by using the name rather than id. Can revert if you don't want it

kosmoz: I see, however it is usually also possible to get logs using compose, e.g. `docker compose logs postgres`. Let's revert it for now please :-)

pmig (approved):

pmig (approved):

PR #1262, opened 2025-10-16
feat(mcp): add streamable http transport method to mcp server

pmig:


Dockerfile.mcp
```diff
@@ -0,0 +1,9 @@
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static-debian12:nonroot@sha256:e8a4044e0b4ae4257efa45fc026c0bc30ad320d43bd4c1a7d5271bd241e386d0
+WORKDIR /
+COPY dist/distr-mcp /distr-mcp
+COPY dist/*.spdx.json /usr/local/share/sbom/
```
kosmoz: This is only for the frontend sbom which is only needed for the main docker image
```suggestion
```

the sbom for the binary is generated by the docker build step automatically (since `sbom: true` is set)

cmd/mcp/main.go
```diff
@@ -105,3 +133,18 @@ func clientConfigFromEnv() (*client.Config, error) {
 
 	return client.NewConfig(opts...), nil
 }
+
+func clientConfigFromEnvOptional() (*client.Config, error) {
```
kosmoz: I think this name is misleading. Maybe rename to `clientConfigFromEnvNoToken` or something like that?

PR #1392, opened 2025-11-19
chore: improve markdown css styling

pmig:


frontend/ui/src/styles/markdown.scss
```diff
@@ -81,6 +86,7 @@
     ol {
       @apply space-y-1;
       @apply list-inside;
+      @apply ml-3;
```
kosmoz: best practice is to use start/end instead of left/right
```suggestion
      @apply ms-3;
```

PR #1416, opened 2025-11-20
feat: add stripe webhook handler, checkout endpoint

kosmoz:


internal/handlers/webhook_stripe.go
```diff
@@ -0,0 +1,163 @@
...
+
+		default:
+			log.Info("unhandled stripe event")
+		}
+
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
+func handleStripeSunscription(ctx context.Context, subscription stripe.Subscription) error {
```
pmig:
```suggestion
func handleStripeSubscription(ctx context.Context, subscription stripe.Subscription) error {
```

mise.toml
```diff
@@ -5,6 +5,7 @@ golangci-lint = "2.6.2"
 helm-docs = "1.14.2"
 watchexec = "2.3.2"
 pnpm = "10.23.0"
+stripe = "latest"
```
pmig:
```suggestion
stripe = "1.32.0"
```

PR #1448, opened 2025-11-25
feat: user roles

kosmoz:


frontend/ui/src/app/components/nav-bar/nav-bar.component.html
pmig, line 125: Your own displayed user role will get displayed incorrectly, when changing the user role of another user in the frontend

frontend/ui/src/app/components/users/users.component.html
```diff
@@ -172,6 +217,42 @@ <h2 class="text-xl font-semibold text-gray-900 dark:text-white">
                     </div>
                   }
                 </div>
+                <div class="space-y-2">
+                  <label for="userRole" class="block text-sm font-medium text-gray-900 dark:text-white">
+                    Permissions <span class="text-red-500">*</span>
```
pmig:
```suggestion
                    User role <span class="text-red-500">*</span>
```

frontend/ui/src/app/components/users/users.component.html
```diff
@@ -172,6 +217,42 @@ <h2 class="text-xl font-semibold text-gray-900 dark:text-white">
...
+                  </label>
+                  <select
+                    id="userRole"
+                    formControlName="userRole"
+                    class="bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-blue-500 focus:border-blue-500 block w-full p-2.5 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-500 dark:focus:border-blue-500">
+                    <option value="read_only">Read Only</option>
+                    <option value="read_write">Read/Write</option>
```
pmig:
```suggestion
                    <option value="read_only">Viewer</option>
                    <option value="read_write">User</option>
```

pmig (changes requested):
I'll need to do further testing, but on the first glance it's so nice to see that we have user roles 🎉 .

Additionally I think that we should add a middle ware that ensures that the current jwt is on the latest "version" to mitigate unwanted side effects.

PR #1541, opened 2025-12-10
feat: add deployment status and resource log export

pmig:


internal/handlers/deployments.go
```diff
@@ -396,6 +442,56 @@ func getDeploymentLogsResourcesHandler() http.HandlerFunc {
...
+
+		// Stream log records directly from database to response
+		err := db.GetDeploymentLogRecordsForExport(
+			ctx, deployment.ID, resource, limit,
+			func(record types.DeploymentLogRecord) error {
+				line := fmt.Sprintf("[%s] [%s] %s\n",
+					record.Timestamp.Format(time.RFC3339),
+					record.Severity,
+					record.Body)
+				_, err := w.Write([]byte(line))
```
kosmoz:
```suggestion
				_, err := fmt.Fprintf(w, "[%s] [%s] %s\n",
					record.Timestamp.Format(time.RFC3339),
					record.Severity,
					record.Body)
```

internal/db/deployments.go
```diff
@@ -396,3 +396,64 @@ func GetDeploymentStatus(
...
+			"deploymentRevisionIds": deploymentRevisionIDs,
+			"limit":                 limit,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to query DeploymentRevisionStatus: %w", err)
+	}
+	defer rows.Close()
+
+	for rows.Next() {
```
kosmoz: `pgx.ForEachRow` would be a bit simpler. Also, I think a more modern/idiomatic approach would be to return an `iter.Seq2`. Maybe next time 😄

internal/subscription/global_limits.go
```diff
@@ -23,6 +23,10 @@ const (
 	MaxDeploymentTargetsPerCustomerOrganizationStarter   Limit = 1
 	MaxDeploymentTargetsPerCustomerOrganizationPro       Limit = 3
 	MaxDeploymentTargetsPerCustomerOrganizationUnlimited Limit = Unlimited
+
+	MaxLogExportRowsStarter    Limit = 100
+	MaxLogExportRowsPro        Limit = 10_000
+	MaxLogExportRowsEnterprise Limit = 1_000_000
```
kosmoz, line 29: we don't keep more than 5000

internal/subscription/global_limits.go
```diff
@@ -84,6 +88,23 @@ func GetDeploymentTargetsPerCustomerOrganizationLimit(st types.SubscriptionType)
 	}
 }
 
+func GetLogExportRowsLimit(st types.SubscriptionType) Limit {
+	switch st {
+	case types.SubscriptionTypeCommunity:
+		return MaxLogExportRowsStarter
+	case types.SubscriptionTypeTrial:
+		return MaxLogExportRowsStarter
```
kosmoz: This is inconsistent. Trial is unlimited in almost every other respect

internal/handlers/deployments.go
```diff
@@ -382,6 +385,49 @@ func getDeploymentStatus(w http.ResponseWriter, r *http.Request) {
...
+		// Stream status records directly from database to response
+		err := db.GetDeploymentStatusForExport(
+			ctx, deployment.ID, limit,
+			func(record types.DeploymentRevisionStatus) error {
+				line := fmt.Sprintf("[%s] [%s] %s\n",
+					record.CreatedAt.Format(time.RFC3339),
+					record.Type,
+					record.Message)
+				_, err := w.Write([]byte(line))
+				return err
```
kosmoz:
```suggestion
				_, err := fmt.Fprintf(w, "[%s] [%s] %s\n",
					record.CreatedAt.Format(time.RFC3339),
					record.Type,
					record.Message)
				return err
```

internal/db/deployment_log_record.go
```diff
@@ -145,6 +145,63 @@ func GetDeploymentLogRecords(
...
+			"resource":     resource,
+			"limit":        limit,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("could not query DeploymentLogRecord: %w", err)
+	}
+	defer rows.Close()
+
+	for rows.Next() {
```
kosmoz: see other comment

frontend/ui/src/app/deployments/deployment-status-modal/timeseries-table.component.ts
```diff
@@ -132,6 +156,55 @@ export class TimeseriesTableComponent {
...
+      exportObservable = this.deploymentStatusService.export(deploymentId);
+      filename = `${today}_deployment_status.log`;
+    } else {
+      this.isExporting = false;
+      return;
+    }
+
+    exportObservable.subscribe({
+      next: (blob) => {
+        const url = window.URL.createObjectURL(blob);
```
kosmoz, line 195: It would be really cool if we could also "stream" the download on the client side by using the browsers builtin download manager instead of buffering the blob. I assume this is done because auth would be missing otherwise, but OTOH it would be very easy to add support for authenticating via query param in the backend. A token extractor func already exists for this.

frontend/ui/src/app/deployments/deployment-status-modal/timeseries-table.component.ts
```diff
@@ -85,8 +103,14 @@ export interface TimeseriesSource {
 })
 export class TimeseriesTableComponent {
   public readonly source = input.required<TimeseriesSource>();
+  public readonly deploymentId = input<string | undefined>(undefined);
+  public readonly resource = input<string | undefined>(undefined);
+  public readonly exportType = input<'logs' | 'status' | undefined>(undefined);
```
kosmoz: no need to specify undefined
```suggestion
  public readonly deploymentId = input<string>();
  public readonly resource = input<string>();
  public readonly exportType = input<'logs' | 'status'>();
```

any chance at least some of these could be required?

frontend/ui/src/app/deployments/deployment-status-modal/timeseries-table.component.ts
```diff
@@ -52,14 +55,29 @@ export interface TimeseriesSource {
...
 
-      @if (hasMore) {
-        <div class="flex items-center justify-center mt-2">
-          <button
-            type="button"
-            class="py-2 px-3 flex items-center text-sm font-medium text-center text-gray-900 focus:outline-none bg-white rounded-lg border border-gray-200 hover:bg-gray-100 hover:text-primary-700 focus:z-10 focus:ring-4 focus:ring-gray-200 dark:focus:ring-gray-700 dark:bg-gray-800 dark:text-gray-400 dark:border-gray-600 dark:hover:text-white dark:hover:bg-gray-700"
-            (click)="showMore()">
-            Load more
-          </button>
+      @if (hasMore || (deploymentId() && exportType())) {
```
kosmoz, line 58: would it make sense to put the button at the top? I found it hard to find

frontend/ui/src/app/deployments/deployment-status-modal/timeseries-table.component.ts
```diff
@@ -20,6 +22,7 @@ export interface TimeseriesSource {
 
 @Component({
   selector: 'app-timeseries-table',
+  standalone: true,
```
kosmoz: standalone is default
```suggestion
```

internal/subscription/global_limits.go
pmig: good point we should increase it to 10_000 via the env

pmig: and for future enterprise / dedicated installations we can come up with a solution later on

internal/handlers/deployments.go
```diff
@@ -396,6 +432,48 @@ func getDeploymentLogsResourcesHandler() http.HandlerFunc {
...
+		}
+
+		authInfo := auth.Authentication.Require(ctx)
+		org := authInfo.CurrentOrg()
+
+		limit := int(subscription.GetLogExportRowsLimit(org.SubscriptionType))
+
+		filename := fmt.Sprintf("%s_%s.log", time.Now().Format("2006-01-02"), resource)
+
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
```
pmig: refactor to set file download method

frontend/ui/src/app/deployments/deployment-status-modal/timeseries-table.component.ts
pmig: Currently users need to use the show more button multiple times, so i guess the export button on that place is fine.

PR #1557, opened 2025-12-15
chore(sdk/js): remove deprecated deployment item in deployment target

pmig:


package.json
```diff
@@ -9,6 +9,7 @@
     "buildconfig": "node hack/update-frontend-version.js",
     "sentry-upload": "hack/sentry-release.sh",
     "start": "pnpm buildconfig && ng serve",
+    "serve:ui": "ng serve ui",
```
kosmoz: should probably run `pnpm buildconfig` first. Also seems to be the same as `start`?

PR #1573, opened 2025-12-16
feat: application link

pmig:


internal/db/deployments.go
```diff
@@ -125,9 +130,74 @@ func GetDeploymentsForDeploymentTarget(
...
+	return result, nil
+}
+
+func TemplateApplicationLink(link string, envFileData []byte, valuesYaml []byte) (string, error) {
+	if link == "" {
+		return "", nil
+	}
+
+	envMap := make(map[string]string)
+	if len(envFileData) > 0 {
```
kosmoz: Please use `dotenv.UnmarshalBytesWithLookup`, it is the same used by docker and our agent.

EDIT: Also, this solution does not work when a comment is at the end of a line

internal/db/deployments.go
```diff
@@ -125,9 +130,74 @@ func GetDeploymentsForDeploymentTarget(
...
+			deployments[i].EnvFileData,
+			deployments[i].ValuesYaml,
+		)
+		if err != nil {
+			continue
+		}
+		deployments[i].ApplicationLink = templatedLink
+	}
+	return nil
 }
```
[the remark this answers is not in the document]
kosmoz: I agree, would it make sense to set the link to nil if templating fails?

internal/db/deployments.go
```diff
@@ -125,9 +130,74 @@ func GetDeploymentsForDeploymentTarget(
...
+		return "", fmt.Errorf("failed to parse link template: %w", err)
+	}
+
+	var buf bytes.Buffer
+	if err := tmpl.Execute(&buf, data); err != nil {
+		return "", fmt.Errorf("failed to execute link template: %w", err)
+	}
+
+	return buf.String(), nil
+}
```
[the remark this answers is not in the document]
kosmoz: I think this could be a fair point. Especially resource exhaustion could be a problem. I don't think we have to do anything for now, but it's something to keep in mind

internal/types/application_version.go
```diff
@@ -15,6 +15,7 @@ type ApplicationVersion struct {
 	CreatedAt     time.Time      `db:"created_at" json:"createdAt"`
 	ArchivedAt    *time.Time     `db:"archived_at" json:"archivedAt,omitempty"`
 	Name          string         `db:"name" json:"name"`
+	Link          string         `db:"link" json:"link"`
```
kosmoz: I think `LinkTemplate` would be a better name, also in the DB. Then in the deployment we can keep `ApplicationLink` and `ApplicationLinkTemplate` separate and it is clear when something is a template and when it isn't. Also, we can then only expose the already executed `ApplicationLink` via JSON.

internal/db/deployments_test.go
```diff
@@ -0,0 +1,178 @@
+package db_test
+
+import (
+	"testing"
+
+	"github.com/glasskube/distr/internal/db"
+	. "github.com/onsi/gomega"
+)
+
+func TestTemplateApplicationLink(t *testing.T) {
```
kosmoz, line 10: Tests are always good. Maybe we should consider running tests in CI? :^)

frontend/ui/src/app/applications/application-detail.component.html
```diff
@@ -310,6 +310,38 @@ <h3 class="block mb-2 text-lg font-medium text-gray-900 dark:text-white">
                   </div>
                 </div>
 
+                <div class="space-y-4 sm:flex sm:space-x-4 sm:space-y-0 mt-4">
```
kosmoz, line 313: I think it would be better to move the link to the bottom, as I don't think it's more important than the other fields

frontend/ui/src/app/applications/application-detail.component.html
```diff
@@ -310,6 +310,38 @@ <h3 class="block mb-2 text-lg font-medium text-gray-900 dark:text-white">
...
+                      id="applicationVersionLink"
+                      placeholder="https://example.com/release-notes"
+                      class="bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-primary-600 focus:border-primary-600 block w-full p-2.5 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500" />
+                  </div>
+                </div>
+                <div class="text-xs text-gray-500 dark:text-gray-400 mt-1">
+                  You can use the Go template language to get a dynamic link based on
+                  @if (application.type === 'docker') {
+                    <span ngNonBindable
+                      >environment variables (<code>{{ .Env.HOSTNAME }}</code
```
kosmoz, line 334: Small suggestion: maybe use the same style as defined in markdown.scss for this `code`? Something like this I think: `whitespace-pre rounded-sm px-1 text-gray-900 dark:text-gray-200 bg-gray-200 dark:bg-gray-800 border border-gray-200 dark:border-gray-600`

frontend/ui/src/app/applications/application-version-detail-modal.component.html
```diff
@@ -0,0 +1,134 @@
...
+        </button>
+      </div>
+
+      <!-- Modal body -->
+      <div class="p-4 md:p-5 space-y-4">
+        @if (versionDetail(); as detail) {
+          <div class="flex flex-wrap col-span-2 text-gray-500 dark:text-gray-400">
+            <!-- Link Tab (Common for both Docker and Kubernetes) -->
+            <input type="radio" id="tab-link" name="version-detail-tabs" class="hidden peer/link" checked />
+            <label
```
kosmoz, line 33: I would change this to `tab-general` or something like that, as having a tab only for the link doesn't make so much sense IMO

frontend/ui/src/app/applications/application-detail.component.html
pmig: I also thought about it, but if there are long templates ppl. will probably not find it :/

internal/db/deployments.go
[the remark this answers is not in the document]
pmig: solved via separate attributes

frontend/ui/src/app/applications/application-detail.component.html
pmig: I tried it, it doesn't look good.

kosmoz (approved):

PR #1638, opened 2026-01-14
fix: handle delayed initial license use case 

pmig:


frontend/ui/src/app/deployments/deployment-form/deployment-form.component.ts
```diff
@@ -175,6 +176,11 @@ export class DeploymentFormComponent implements OnInit, AfterViewInit, OnDestroy
     shareReplay(1)
   );
 
+  /**
+   * If the license is not initially set, this update must be sent to the server.
+   */
```
kosmoz: This comment is confusing, I would remove it

frontend/ui/src/app/deployments/deployment-form/deployment-form.component.ts
```diff
@@ -175,6 +176,11 @@ export class DeploymentFormComponent implements OnInit, AfterViewInit, OnDestroy
...
+  protected readonly licenseUpdateRequired$ = new BehaviorSubject<boolean>(false);
```
kosmoz, line 182: Note that a `BehaviorSubject` is almost never a good solution and I think this case is no exception. However, a better solution would require knowledge of the existing deployment state which this component currently lacks. 

So, to avoid bigger changes, I think we can let it pass this time but I thought it is worth mentioning.

internal/db/deployments.go
```diff
@@ -248,6 +248,33 @@ func UpdateDeployment(ctx context.Context, deployment *types.Deployment) error {
...
+		SET application_license_id = @applicationLicenseID
+		WHERE id = @id
+		RETURNING`+deploymentOutputExpr,
+		pgx.NamedArgs{
+			"id":                   deployment.ID,
+			"applicationLicenseID": deployment.ApplicationLicenseID,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("could not update Deployment license: %w", err)
```
kosmoz: our convention is to use the entity name in the error message

```suggestion
		return fmt.Errorf("could not update Deployment: %w", err)
```

internal/db/deployments.go
```diff
@@ -248,6 +248,33 @@ func UpdateDeployment(ctx context.Context, deployment *types.Deployment) error {
...
+	}
+	if result, err := pgx.CollectExactlyOneRow(rows, pgx.RowToStructByName[types.Deployment]); err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			err = apierrors.ErrNotFound
+		}
+		return fmt.Errorf("could not update Deployment license: %w", err)
```
kosmoz:
```suggestion
		return fmt.Errorf("could not update Deployment: %w", err)
```

internal/handlers/deployments.go
```diff
@@ -114,6 +114,26 @@ func putDeployment(w http.ResponseWriter, r *http.Request) {
...
+				*deploymentRequest.DeploymentID,
+				authInfo.CurrentUserID(),
+				*authInfo.CurrentOrgID(),
+				authInfo.CurrentCustomerOrgID(),
+			)
+			if err == nil && deployment != nil && deployment.ApplicationLicenseID == nil &&
+				deploymentRequest.ApplicationLicenseID != nil {
+				deployment.ApplicationLicenseID = deploymentRequest.ApplicationLicenseID
+
+				if err := db.UpdateDeployment(ctx, deployment); err != nil {
```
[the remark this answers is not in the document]
kosmoz: this is correct

internal/handlers/deployments.go
```diff
@@ -114,6 +114,26 @@ func putDeployment(w http.ResponseWriter, r *http.Request) {
...
+					log.Warn("could not set initial license for deployment", zap.Error(err))
+					sentry.GetHubFromContext(ctx).CaptureException(err)
+					http.Error(w, err.Error(), http.StatusInternalServerError)
+					return err
+				}
+			}
```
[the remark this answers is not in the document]
kosmoz: this is also correct. `err == nil` should be removed above and instead there should be a `if err != nil { }` that returns early.

internal/handlers/deployments.go
```diff
@@ -264,10 +284,12 @@ func validateDeploymentRequest(
 				request.ApplicationLicenseID = existingDeployment.ApplicationLicenseID
 			}
 		} else if existingDeployment.ApplicationLicenseID == nil {
-			return badRequestError(w, "can not update license")
+			// initially setting a license after a first license required license configuration and can be set once
+			// continue
```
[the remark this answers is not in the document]
kosmoz: I agree that this comment is unclear

internal/handlers/deployments.go
```diff
@@ -114,6 +114,26 @@ func putDeployment(w http.ResponseWriter, r *http.Request) {
...
+				authInfo.CurrentUserID(),
+				*authInfo.CurrentOrgID(),
+				authInfo.CurrentCustomerOrgID(),
+			)
+			if err == nil && deployment != nil && deployment.ApplicationLicenseID == nil &&
+				deploymentRequest.ApplicationLicenseID != nil {
+				deployment.ApplicationLicenseID = deploymentRequest.ApplicationLicenseID
+
+				if err := db.UpdateDeployment(ctx, deployment); err != nil {
+					log.Warn("could not set initial license for deployment", zap.Error(err))
```
kosmoz: the deployment already exists, so "initial" is not the right word IMO

```suggestion
					log.Warn("could not set license for deployment", zap.Error(err))
```

kosmoz (changes requested):

PR #1639, opened 2026-01-14
chore: add dlv to mise for debugging

pmig:


mise.toml
```diff
@@ -6,6 +6,7 @@ helm-docs = "1.14.2"
 watchexec = "2.3.2"
 pnpm = "10.28.0"
 stripe = "1.34.0"
+"go:github.com/go-delve/delve/cmd/dlv" = "1.25.2"
```
kosmoz: I think we should use the latest version. I just tried it and it works.

```suggestion
"go:github.com/go-delve/delve/cmd/dlv" = "1.26.0"
```

PR #1641, opened 2026-01-14
chore(ui): hide tutorial navlink if org has subscription

kosmoz:


frontend/ui/src/app/components/side-bar/side-bar.component.ts
```diff
@@ -76,6 +80,22 @@ export class SideBarComponent {
...
+    this.organizationService
+      .get()
+      .pipe(
+        map(
+          (org) =>
+            !(
+              org.subscriptionType === 'starter' ||
+              org.subscriptionType === 'pro' ||
+              org.subscriptionType === 'enterprise'
+            )
```
pmig, line 93:
```suggestion
          (org) => org.subscriptionType === 'trial'
```

Or am I missing something?

The same computation is already don in the nav-shell component and could be defined as an input.

pmig (changes requested):

frontend/ui/src/app/components/side-bar/side-bar.component.ts
kosmoz: No, there is 'community' as well. I think a grouping like this is closer to the semantic of "hide if organization is on a paid plan", that's why I chose to do it this way.

pmig: gotcha, I am switching the logic for the "contact founders" to the same logic, I think that we want to show that option to our community users.

PR #1664, opened 2026-01-16
feat: add deployment target notes

kosmoz:


frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.ts
```diff
@@ -224,6 +238,27 @@ export class DeploymentTargetCardComponent {
...
+
+    if (!id) {
+      return;
+    }
+
+    this.notesFormLoading = true;
+    this.deploymentTargets.saveNotes(id, notes).subscribe({
+      next: () => {
+        this.toast.success('Notes saved successfully');
+        this.notesFormLoading = false;
```
pmig, line 253: This toast is below the sidebar, so it is not visible to users, it should be above the drawer

pmig (changes requested):

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.ts
kosmoz: this has been true before, it's not a regression in this PR

pmig: Other sidebars close on submit, so the toast will be visible. So I think that we should tackle this in this pr.

kosmoz: see e.g. application license sidebar when there is an error. I will do it separately

kosmoz: https://github.com/distr-sh/distr/pull/1675

pmig (approved):

PR #1690, opened 2026-01-21
feat: add customer organization to artifact version pull

pmig:


internal/handlers/artifact_pulls.go
```diff
@@ -61,6 +61,28 @@ func getArtifactPullsHandler() http.HandlerFunc {
...
+			if pull.UserAccount != nil {
+				if pull.UserAccount.Name != "" {
+					response[i].UserAccountName = &pull.UserAccount.Name
+				}
+				response[i].UserAccountEmail = &pull.UserAccount.Email
+			}
+
+			if pull.CustomerOrganization != nil {
+				response[i].CustomerOrganizationName = &pull.CustomerOrganization.Name
+			}
```
kosmoz: we could also map the full user and customer org, that would allow us to show avatars in the UI. 

In any case, please move this code to `internal/mapping` and use `mapping.List(...)`

PR #1808, opened 2026-02-11
feat: artifact pulls filters

pmig:


internal/db/artifacts.go
```diff
@@ -690,37 +689,159 @@ func EnsureArtifactTagLimitForInsert(ctx context.Context, orgID uuid.UUID) (bool
...
 
-func GetArtifactVersionPulls(
+func GetArtifactVersionPullFilterOptions(
+	ctx context.Context,
+	orgID uuid.UUID,
+) (*types.ArtifactVersionPullFilterOptions, error) {
+	db := internalctx.GetDb(ctx)
+	result := &types.ArtifactVersionPullFilterOptions{}
+
+	pullBaseJoin := `FROM ArtifactVersionPull p
```
kosmoz: you could remove the WHERE part from this common expression. that would allow to use it in 5/5 queries instead of only 2/5

internal/db/artifacts.go
```diff
@@ -690,37 +689,159 @@ func EnsureArtifactTagLimitForInsert(ctx context.Context, orgID uuid.UUID) (bool
 	}
 }
 
-func GetArtifactVersionPulls(
+func GetArtifactVersionPullFilterOptions(
```
kosmoz, line 692: We should absolutely performance test this before merging. There are several SELECT DISTINCT queries on a huge table which are notorious for being slow. GROUP BY would probably be better but we should still test!

internal/handlers/artifact_pulls.go
```diff
@@ -26,36 +32,123 @@ func ArtifactPullsRouter(r chiopenapi.Router) {
...
+		filter.UserAccountID = &id
+	}
+
+	if addr, err := QueryParam(r, "remoteAddress", parseString); errors.Is(err, ErrParamNotDefined) {
+		// use default
+	} else if err != nil {
+		return filter, err
+	} else {
+		filter.RemoteAddress = &addr
+	}
```
kosmoz: I would honestly change this to 
```suggestion
	if addr := r.FormValue("remoteAddress"); addr != "" 
		filter.RemoteAddress = &addr
	}
```
much simpler

internal/handlers/artifact_pulls.go
```diff
@@ -66,3 +159,109 @@ func getArtifactPullsHandler() http.HandlerFunc {
...
+		}
+
+		csvWriter.Flush()
+		if err := csvWriter.Error(); err != nil {
+			log.Warn("CSV flush error", zap.Error(err))
+		}
+	}
+}
+
+func ptrOrEmpty(s *string) string {
```
kosmoz: I think "derefOrDefault" would be a more correct name.

api/artifact.go
```diff
@@ -25,3 +25,15 @@ type ArtifactVersionPullResponse struct {
 	Artifact                 types.Artifact        `json:"artifact"`
 	ArtifactVersion          types.ArtifactVersion `json:"artifactVersion"`
 }
+
+type ArtifactPullFilterOption struct {
+	ID   string `json:"id"`
```
kosmoz: should be uuid.UUID

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -1,48 +1,223 @@
...
+
+  protected readonly customerOrgFilter = new FormControl('');
+  protected readonly userAccountFilter = new FormControl('');
+  protected readonly remoteAddressFilter = new FormControl('');
+  protected readonly artifactFilter = new FormControl('');
+  protected readonly artifactVersionFilter = new FormControl('');
+  protected readonly dateFromFilter = new FormControl('');
+  protected readonly dateToFilter = new FormControl('');
+
+  private readonly filterDefs: FilterDef[] = [
```
kosmoz: Why is this not a FormGroup?

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -1,48 +1,223 @@
...
+    const queryParams: Params = {};
+    for (let i = 0; i < this.filterDefs.length; i++) {
+      queryParams[this.filterDefs[i].queryParam] = values[i] || null;
+    }
+    this.router.navigate([], {queryParams, replaceUrl: true});
+  }
+
+  private buildFilters(values: (string | null)[]): ArtifactPullFilters {
+    const filters: ArtifactPullFilters = {};
+    const [custOrg, user, addr, artifact, version, from, to] = values;
```
kosmoz: This order is duplicated all over the place. I think this will be hard to manage if we inevitably want to change things. Please use a FormGroup if possible or maybe at least use an object literal to make it order independent

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -1,48 +1,223 @@
...
+    {queryParam: 'customerOrganizationId', control: this.customerOrgFilter},
+    {queryParam: 'userAccountId', control: this.userAccountFilter},
+    {queryParam: 'remoteAddress', control: this.remoteAddressFilter},
+    {queryParam: 'artifactId', control: this.artifactFilter},
+    {queryParam: 'artifactVersionId', control: this.artifactVersionFilter},
+    {queryParam: 'from', control: this.dateFromFilter},
+    {queryParam: 'to', control: this.dateToFilter},
+  ];
+
+  protected readonly filterOptions$ = this.pullsService.getFilterOptions().pipe(shareReplay(1));
```
kosmoz: if possible, I think we should try to use signals as much as possible instead of shareReplay and AsyncPipe

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -1,48 +1,223 @@
...
+    this.pullsService.export(filters).subscribe({
+      next: (blob) => {
+        const url = window.URL.createObjectURL(blob);
+        const a = document.createElement('a');
+        a.href = url;
+        a.download = `${new Date().toISOString().split('T')[0]}_artifact_pulls.csv`;
+        document.body.appendChild(a);
+        a.click();
+        window.URL.revokeObjectURL(url);
+        document.body.removeChild(a);
```
kosmoz: this is exactly the same as in TimeseriesTableComponent. maybe refactor to a util

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -1,48 +1,223 @@
...
+      filters.artifactVersionId = version;
+    }
+    if (from) {
+      const [y, m, d] = from.split('-').map(Number);
+      filters.after = new Date(y, m - 1, d);
+    }
+    if (to) {
+      const [y, m, d] = to.split('-').map(Number);
+      filters.before = new Date(y, m - 1, d, 23, 59, 59, 999);
+    }
```
kosmoz, line 211: This seems unnecessarily complex. `new Date(string)` works just fine. Or use `dayjs` which would also make the second part (add one day) easier

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -1,48 +1,223 @@
...
   protected formatRemoteAddress(addr: string): string {
     if (addr.includes(']')) {
-      // IPv6
       return addr.substring(0, addr.lastIndexOf(']') + 1);
     } else if (addr.includes(':')) {
-      // IPv4
       return addr.substring(0, addr.lastIndexOf(':'));
     } else {
-      // fallback for undetermined format
       return addr;
```
kosmoz, line 45: why remove these comments? I actually think the code is not very clear without them

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -1,48 +1,223 @@
...
+        toastRef.toastRef.close();
+        this.toast.error('Export failed');
+      },
+    });
+  }
+
+  protected formatVersionName(name: string): string {
+    const shaPrefix = 'sha256:';
+    if (name.startsWith(shaPrefix)) {
+      return shaPrefix + name.substring(shaPrefix.length, shaPrefix.length + 8);
```
kosmoz: On the artifact versions page we truncate to 10 characters after the prefix.

Also, I think this would be simpler:
```suggestion
      return name.substring(0, shaPrefix.length + 10);
```

internal/db/artifacts.go
```diff
@@ -690,37 +689,160 @@ func EnsureArtifactTagLimitForInsert(ctx context.Context, orgID uuid.UUID) (bool
...
+		conditions = append(conditions, "p.customer_organization_id = @customerOrgId")
+		args["customerOrgId"] = *filter.CustomerOrganizationID
+	}
+	if filter.UserAccountID != nil {
+		conditions = append(conditions, "p.useraccount_id = @userAccountId")
+		args["userAccountId"] = *filter.UserAccountID
+	}
+	if filter.RemoteAddress != nil {
+		conditions = append(conditions, "p.remote_address = @remoteAddress")
+		args["remoteAddress"] = *filter.RemoteAddress
```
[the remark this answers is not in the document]
kosmoz: This is a fair point

cmd/hub/generate/pulls/generate.go
```diff
@@ -0,0 +1,108 @@
...
+	for inserted < totalCount {
+		remaining := totalCount - inserted
+		currentBatch := batchSize
+		if remaining < currentBatch {
+			currentBatch = remaining
+		}
+
+		count, err := db.CopyFrom(
+			ctx,
+			pgx.Identifier{"artifactversionpull"},
```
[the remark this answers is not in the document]
kosmoz: copilot brings this up every time but it is wrong. pgx.Identifier only works with lowercase!

PR #1823, opened 2026-02-12
chore(backend): allow application license version narrowing

pmig:


internal/handlers/application_licenses.go
```diff
@@ -162,21 +188,12 @@ func updateApplicationLicense(w http.ResponseWriter, r *http.Request) {
 				return v.ID == existingVersion.ID
 			})
 			if !stillExists {
-				if len(license.Versions) > 0 {
-					// for now, removing specific versions from the license is not possible
-					// for removal we also would have to check whether this version is used in some deployment target
-					err = errors.New("narrowing down license scope is not allowed yet")
-					http.Error(w, err.Error(), http.StatusBadRequest)
+				if err := db.RemoveVersionFromApplicationLicense(
```
kosmoz: since we're already touching it, I think this might be a good opportunity to simplify this logic and let postgres do some of it. Here is an example of how this can also be done (should be more performant too): 

https://github.com/distr-sh/distr/blob/70edce8bb432b7b9337d5429cac76d9dabaeb7ce/internal/db/alert_configuration.go#L245-L287

kosmoz (changes requested):
I get an error trying to update a license

> could not collect ApplicationLicense: can't scan into dest[10] (col: versions): failed to scan array element 0: Scan: invalid UUID length: 3

Issue #1814, opened 2026-02-11
OCI registry returns 401 for agent after repeated helm values updates on a deployment


robbeverhelst:
## Description

After repeatedly updating the helm values for a specific artifact on a deployment target via the vendor portal UI, the agent starts receiving `401 Unauthorized` from the Distr OCI registry when pulling that artifact. Other artifacts assigned to the same deployment target continue to authenticate and pull successfully.

## Steps to reproduce

1. Create a deployment target with multiple artifacts (e.g. artifact-a, artifact-b, artifact-c)
2. All artifacts deploy successfully on first install
3. Update artifact-c's helm values multiple times via the vendor portal
4. After several updates, the agent fails to pull artifact-c with 401
5. artifact-a and artifact-b continue working on the same agent

## Agent error

```
upgrade error: helm upgrade failed: helm preflight failed: could not locate chart: 
failed to perform "FetchReference" on source: 
GET "https://<distr-host>/v2/<org>/<artifact>/manifests/<version>": 
response status code 401: Unauthorized
```

## Server-side registry logs

```
GET /v2/ → 401  (token for affected artifact)
GET /v2/ → 200  (token for other artifacts on same target)
GET /v2/<org>/<artifact>/manifests/<version> → 401
```

## Does NOT fix it

- Removing the artifact from the deployment and re-adding it
- Undeploying and redeploying the artifact

## Does fix it

- Creating a new deployment target with fresh agent credentials

## Environment

- Distr version: 2.9.0 CE
- Internal (self-deployment) setup. Unclear if it also affects customer deployments.

## Suspected cause

The per-artifact registry auth token gets invalidated after repeated values updates but isn't properly regenerated when the artifact is removed/re-added to the deployment.

---

We can provide additional info (server-side logs, agent logs, deployment IDs) if that helps narrow down the root cause. Let us know what would be useful.

kosmoz:
Hi @robbeverhelst, thank you for reporting.

I'm having some troubles reproducing this bug, maybe you can help me out a bit?

* If possible, could you share some info about the charts you are using? Do you install them via tag or digest?
* Could you also share some more of the agent/server logs?
* Did you change the application version between deploys? Or did you only change the helm values?

Aside: The Agent doesn't use different authentication tokens per artifact. There is a single token used for all API and registry requests so this behavior is a bit puzzling…

PR #1875, opened 2026-02-20
chore(registry): more granular "access denied" messages

kosmoz:


pmig:
this change provides additional insights into https://github.com/distr-sh/distr/issues/1814

PR #1910, opened 2026-02-26
feat: support bundels

pmig:
- [x] Inform the user what happens when showing the curl script


internal/migrations/sql/84_support_bundle_tables.up.sql
```diff
@@ -0,0 +1,60 @@
+ALTER TYPE CUSTOMER_ORGANIZATION_FEATURE ADD VALUE IF NOT EXISTS 'support_bundles';
+
+CREATE TABLE SupportBundleConfiguration (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    created_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+    organization_id UUID NOT NULL UNIQUE REFERENCES Organization (id) ON DELETE CASCADE
+);
+
+CREATE TABLE SupportBundleConfigurationEnvVar (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
kosmoz: do we need an ID here? could we not just match by support_bundle_configuration_id and name?

internal/resources/embedded/support-bundle/collect-script.sh
kosmoz, line 1: I would move this to the resources package

internal/migrations/sql/84_support_bundle_tables.up.sql
```diff
@@ -0,0 +1,60 @@
+ALTER TYPE CUSTOMER_ORGANIZATION_FEATURE ADD VALUE IF NOT EXISTS 'support_bundles';
+
+CREATE TABLE SupportBundleConfiguration (
```
kosmoz: this table has no purpose

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+					api.CreateSupportBundleCommentRequest
+				}{})).
+				With(option.Response(http.StatusOK, api.SupportBundleComment{}))
+		})
+	})
+}
+
+// SupportBundleScriptRouter handles endpoints called by the collect script.
+// The collect-script endpoint uses query-param auth; the other endpoints use PAT header auth.
+func SupportBundleScriptRouter(r chiopenapi.Router) {
```
kosmoz: since this file is so huge, I think this function and the related handlers should be moved to a new file

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+		}
+
+		if request.Name == "" || request.Content == "" {
+			http.Error(w, "name and content are required", http.StatusBadRequest)
+			return
+		}
+
+		const maxContentSize = 5 * 1024 * 1024 // 5MB
+		if len(request.Content) > maxContentSize {
+			http.Error(w, "content exceeds maximum size of 5MB", http.StatusBadRequest)
```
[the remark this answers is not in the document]
kosmoz: this is true

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+		}
+
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		if _, err := w.Write([]byte(script)); err != nil {
+			log.Warn("failed to write collect script", zap.Error(err))
+		}
+	}
+}
+
+func uploadSupportBundleResourceHandler() http.HandlerFunc {
```
kosmoz: maybe this endpoint should be multipart/formdata

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+		if err != nil {
+			if errors.Is(err, apierrors.ErrNotFound) {
+				http.Error(w, "invalid or expired token", http.StatusUnauthorized)
+			} else {
+				log.Error("failed to validate support bundle token", zap.Error(err))
+				sentry.GetHubFromContext(ctx).CaptureException(err)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
+			return
+		}
```
kosmoz: you can use `auth.Authentication` for this if you add a FromQuery token extractor

frontend/ui/src/app/components/dashboard/dashboard.component.html
```diff
@@ -1,5 +1,23 @@
 <div class="p-4 bg-gray-50 dark:bg-gray-900">
   <div class="p-4">
+    @if (supportBundlesByCustomer$ | async; as supportBundlesByCustomer) {
```
kosmoz: I think this would be a good opportunity to migrate this component to use signals instead of async pipe

frontend/ui/src/app/components/dashboard/dashboard.component.ts
```diff
@@ -21,7 +29,26 @@ export class DashboardComponent implements OnInit, OnDestroy {
   private readonly router = inject(Router);
   private readonly toast = inject(ToastService);
   private readonly dashboardService = inject(DashboardService);
+  private readonly supportBundlesService = inject(SupportBundlesService);
   protected readonly artifactsByCustomer$ = this.dashboardService.getArtifactsByCustomer().pipe(shareReplay(1));
+  protected readonly supportBundlesByCustomer$ = this.supportBundlesService.list().pipe(
+    map((bundles) => {
+      const grouped = new Map<string, SupportBundle[]>();
+      for (const bundle of bundles) {
+        const existing = grouped.get(bundle.customerOrganizationName);
```
kosmoz: edge case: the customer name is not unique

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundles.component.ts
```diff
@@ -0,0 +1,186 @@
...
+  name: FormControl<string>;
+  redacted: FormControl<boolean>;
+}>;
+
+@Component({
+  selector: 'app-vendor-support-bundles',
+  templateUrl: './vendor-support-bundles.component.html',
+  imports: [ReactiveFormsModule, FaIconComponent],
+})
+export class VendorSupportBundlesComponent {
```
kosmoz: This component is poorly named, I couldn't find it because I was looking for something with "settings" in it

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundles.component.ts
```diff
@@ -0,0 +1,186 @@
...
+  private readonly toast = inject(ToastService);
+  private readonly overlay = inject(OverlayService);
+
+  protected readonly loading = signal(true);
+  protected readonly saving = signal(false);
+  protected readonly configExists = signal(false);
+
+  protected readonly envVarsArray = new FormArray<EnvVarFormGroup>([]);
+  private readonly envVarVersion = signal(0);
+  protected readonly duplicateIndices = computed(() => {
```
kosmoz: It would be much better to make this a validator on the FormArray. Then the ugly envVarVersion hack would not be needed

frontend/ui/src/app/components/side-bar/side-bar.component.html
```diff
@@ -219,6 +219,20 @@
           </a>
         </li>
 
+        <li>
```
kosmoz, line 197: should this be a pro feature?

frontend/ui/src/app/support-bundles/customer/customer-support-bundles.component.html
```diff
@@ -0,0 +1,139 @@
...
+            class="border-t border-gray-200 dark:border-gray-600 text-xs text-gray-700 uppercase bg-gray-100 dark:bg-gray-700 dark:text-gray-400">
+            <tr>
+              <th scope="col" class="p-4">Title</th>
+              <th scope="col" class="p-4">Status</th>
+              <th scope="col" class="p-4">Created At</th>
+              <th scope="col" class="p-4">Resources</th>
+            </tr>
+          </thead>
+          <tbody>
+            @for (bundle of bundles$ | async; track bundle.id) {
```
kosmoz: should use toSignal instead of async pipe

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundles.component.html
```diff
@@ -0,0 +1,154 @@
...
+            <p class="text-sm text-gray-500 dark:text-gray-400">
+              No environment variables configured. Click "Add Variable" to add one.
+            </p>
+          }
+
+          <div class="flex justify-center w-full space-x-4 mt-8 sm:mt-0">
+            <button
+              type="button"
+              (click)="save()"
+              [disabled]="saving() || duplicateIndices().size > 0"
```
kosmoz:
```suggestion
              [disabled]="saving() || duplicateIndices().size > 0 || envVarsArray.pristine"
```

and also mark the form as pristine after save

frontend/ui/src/app/support-bundles/vendor/support-bundle-settings.component.html
```diff
@@ -0,0 +1,154 @@
...
+              </button>
+            </div>
+          </div>
+
+          @if (envVarsArray.length > 0) {
+            <div class="space-y-3">
+              @for (group of envVarsArray.controls; track $index) {
+                <div class="flex items-center gap-4">
+                  <div class="flex-1">
+                    <input
```
kosmoz, line 49: maybe add autotrim

api/support_bundle.go
```diff
@@ -0,0 +1,117 @@
...
+		if !envVarNamePattern.MatchString(name) {
+			return validation.NewValidationFailedError(
+				fmt.Sprintf("invalid environment variable name: %v (must match [A-Za-z_][A-Za-z0-9_]*)", ev.Name))
+		}
+		key := strings.ToLower(name)
+		if _, exists := seen[key]; exists {
+			return validation.NewValidationFailedError(
+				fmt.Sprintf("duplicate environment variable name: %v", ev.Name))
+		}
+		seen[key] = struct{}{}
```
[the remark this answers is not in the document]
kosmoz: This is true, it is currently possible to save env vars that have spaces at the beginning or end

frontend/ui/src/app/support-bundles/customer/customer-support-bundles.component.ts
```diff
@@ -0,0 +1,83 @@
...
+import {AutotrimDirective} from '../../directives/autotrim.directive';
+import {DialogRef, OverlayService} from '../../services/overlay.service';
+import {SupportBundlesService} from '../../services/support-bundles.service';
+import {ToastService} from '../../services/toast.service';
+
+@Component({
+  selector: 'app-customer-support-bundles',
+  templateUrl: './customer-support-bundles.component.html',
+  imports: [AsyncPipe, DatePipe, ReactiveFormsModule, RouterLink, FaIconComponent, ClipComponent, AutotrimDirective],
+  animations: [drawerFlyInOut],
```
kosmoz: these animations are deprecated. please only use the new animations in new code. https://angular.dev/guide/animations

We don't have keyframes for the side drawer yet but they are usually easy to convert from the old angular syntax

frontend/ui/src/app/types/support-bundle.ts
```diff
@@ -0,0 +1,67 @@
...
+  createdAt: string;
+  customerOrganizationId: string;
+  customerOrganizationName: string;
+  createdByUserAccountId: string;
+  createdByUserName: string;
+  createdByImageUrl?: string;
+  title?: string;
+  description?: string;
+  status: 'initialized' | 'created' | 'resolved';
+  resourceCount: number;
```
kosmoz, line 23: maybe also add comments count?

frontend/ui/src/app/support-bundles/vendor/support-bundle-settings.component.html
kosmoz, line 1: this page needs a back link

kosmoz, line 1: super admins can edit the env vars on this page. this should be prevented (also in the backend!)

frontend/ui/src/app/support-bundles/customer/customer-support-bundles.component.html
```diff
@@ -0,0 +1,139 @@
...
+          <code class="text-sm text-gray-900 dark:text-white break-all flex-1 select-all">{{ cmd }}</code>
+          <app-clip [clip]="cmd" />
+        </div>
+      </div>
+    } @else {
+      <form [formGroup]="createForm" (ngSubmit)="createBundle()">
+        <div class="space-y-4">
+          <div>
+            <label for="title" class="block mb-2 text-sm font-medium text-gray-900 dark:text-white">Title</label>
+            <input
```
kosmoz: Title should be mandatory IMO. Otherwise, once there are two records without a name everyone is going to be confused.

frontend/ui/src/app/support-bundles/customer/customer-support-bundles.component.html
```diff
@@ -0,0 +1,139 @@
...
+              </tr>
+            }
+          </tbody>
+        </table>
+      </div>
+    </div>
+  </div>
+</section>
+
+<ng-template #createBundleDrawer>
```
kosmoz: Personally, i would move this to a dialog. It's what we normally use especially for this kind of multi-step flow. Anecdotally, I when trying it for the first time I just dismissed the drawer after creation. Because I didn't know that the curl command is shown in the drawer I was unable to do anything with the support bundle. 

On a related note: I think it would make sense to allow users to show the support bundle command again after creation

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,137 @@
...
+        <h2 class="text-lg font-semibold text-gray-900 dark:text-white mb-4">
+          Resources ({{ bundle.resources.length }})
+        </h2>
+        @if (bundle.resources.length === 0) {
+          <p class="text-sm text-gray-500 dark:text-gray-400">No resources collected yet.</p>
+        } @else {
+          <div class="space-y-3">
+            @for (resource of bundle.resources; track resource.id) {
+              <div class="border border-gray-200 dark:border-gray-700 rounded-lg">
+                <button
```
kosmoz, line 91: Hover bug:

Image

frontend/ui/src/app/support-bundles/vendor/support-bundle-settings.component.html
```diff
@@ -0,0 +1,154 @@
+<section class="bg-gray-50 dark:bg-gray-900 p-3 sm:p-5 antialiased">
+  <div class="mx-auto max-w-screen-lg px-4">
+    <div class="bg-white dark:bg-gray-800 relative shadow-md rounded-lg overflow-hidden p-10">
+      <h1
+        class="pb-4 mb-4 text-3xl font-extrabold leading-none tracking-tight text-gray-900 md:text-4xl dark:text-white border-b border-gray-200 dark:border-gray-700">
+        Docker Support Bundles
+      </h1>
+      <p class="text-gray-500 dark:text-gray-400">
+        Configure the environment variables that should be collected from Docker containers when customers create
```
kosmoz: Either the script is bugged or this text is incorrect. At least when I tried it, the script would collect ALL variables from containers and additionally the ones specified on the docker host. This is unexpected…

Redacting works as expected.

internal/resources/embedded/support-bundle/collect-script.sh
kosmoz, line 1: I think users would like to see what has been collected before sending it off. At least system info and env variables

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,710 @@
...
+
+		key, err := authkey.NewKey()
+		if err != nil {
+			log.Error("failed to generate PAT key", zap.Error(err))
+			sentry.GetHubFromContext(ctx).CaptureException(err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		expiresAt := time.Now().Add(1 * time.Hour)
```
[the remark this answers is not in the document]
kosmoz: I agree that this is very short

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+		token := types.AccessToken{
```
kosmoz: Using access tokens for this is kind of an ugly hack. It creates unrelated resources with unclear relation which might lead to confusion, false bug reports or security considerations. A better approach would be storing a salted secret with the support bundle and using that for authentication.

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+			ExpiresAt:      &expiresAt,
+			Label:          util.PtrTo("support-bundle"),
```
kosmoz: This should contain the bundle ID or name

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,137 @@
...
+            </a>
+            <h1 class="text-xl font-semibold text-gray-900 dark:text-white">
+              {{ bundle.title || 'Support Bundle ' + shortId(bundle.id) }}
+            </h1>
+            <span class="ml-3 text-xs font-medium px-2.5 py-0.5 rounded-full" [ngClass]="statusClass(bundle.status)">
+              {{ bundle.status }}
+            </span>
+          </div>
+          @if (auth.isVendor() && bundle.status === 'created') {
+            <button
```
kosmoz: I think this button "pops" too much... maybe change to a grey outline button with green icon only?

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,137 @@
...
+              <fa-icon [icon]="faArrowLeft"></fa-icon>
+            </a>
+            <h1 class="text-xl font-semibold text-gray-900 dark:text-white">
+              {{ bundle.title || 'Support Bundle ' + shortId(bundle.id) }}
+            </h1>
+            <span class="ml-3 text-xs font-medium px-2.5 py-0.5 rounded-full" [ngClass]="statusClass(bundle.status)">
+              {{ bundle.status }}
+            </span>
+          </div>
+          @if (auth.isVendor() && bundle.status === 'created') {
```
kosmoz: missing permission check (also in the backend)

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,137 @@
...
+                  </div>
+                  <p class="text-sm text-gray-700 dark:text-gray-300 whitespace-pre-wrap">{{ comment.content }}</p>
+                </div>
+              </div>
+            }
+          </div>
+        }
+
+        <!-- New comment form -->
+        <form
```
kosmoz: missing permission check (also in the backend)

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+			return
+		}
+
+		status := types.SupportBundleStatus(request.Status)
+		if status != types.SupportBundleStatusResolved {
+			http.Error(w, "only 'resolved' status is allowed", http.StatusBadRequest)
+			return
+		}
+
+		if err := db.UpdateSupportBundleStatus(ctx, id, *auth.CurrentOrgID(), status); errors.Is(err, apierrors.ErrNotFound) {
```
kosmoz: would it make sense to also save who updated the status?

frontend/ui/src/app/services/support-bundles.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    return this.httpClient.patch<SupportBundle>(`${baseUrl}/${id}/status`, request);
+  }
+
+  public getResources(bundleId: string) {
+    return this.httpClient.get<SupportBundleResource[]>(`${baseUrl}/${bundleId}/resources`);
+  }
+
+  public getComments(bundleId: string) {
+    return this.httpClient.get<SupportBundleComment[]>(`${baseUrl}/${bundleId}/comments`);
+  }
```
kosmoz: unused

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+
+			r.Get("/resources", getSupportBundleResourcesHandler()).
+				With(option.Description("List support bundle resources")).
+				With(option.Request(SupportBundleIDRequest{})).
+				With(option.Response(http.StatusOK, []api.SupportBundleResource{}))
+
+			r.Get("/comments", getSupportBundleCommentsHandler()).
+				With(option.Description("List support bundle comments")).
+				With(option.Request(SupportBundleIDRequest{})).
+				With(option.Response(http.StatusOK, []api.SupportBundleComment{}))
```
kosmoz: unused

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+			return
+		}
+
+		commentWithUser, err := db.GetSupportBundleCommentByID(ctx, comment.ID)
+		if err != nil {
+			log.Error("failed to get support bundle comment", zap.Error(err))
+			sentry.GetHubFromContext(ctx).CaptureException(err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
```
kosmoz: this should be one query (`WITH inserted AS (...) SELECT ... FROM inserted JOIN ...`; we have examples for this)

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+
+		RespondJSON(w, mapping.SupportBundleCommentToAPI(*commentWithUser))
+	}
+}
+
+// Script endpoints (called by the collect script)
+
+func queryAuthSupportBundleMiddleware(next http.Handler) http.Handler {
```
kosmoz: This middleware is only used for one endpoint and that endpoint doesn't use any of the data it puts into the context and also seems to do authentication again. I think this middleware might be unnecessary

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,760 @@
...
+			Content:         request.Content,
+		}
+		if err := db.CreateSupportBundleResource(ctx, &resource); err != nil {
+			log.Error("failed to create support bundle resource", zap.Error(err))
+			sentry.GetHubFromContext(ctx).CaptureException(err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		RespondJSON(w, mapping.SupportBundleResourceToAPI(resource))
```
kosmoz: Is it necessary to respond with the full content here again?

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
pmig: Should read only users be able to collect support bundles and create comments on support bundles?

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
pmig: see: https://github.com/distr-sh/distr/pull/1910#discussion_r2877762206

internal/handlers/support_bundles.go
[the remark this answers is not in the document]
pmig: no need to change as this is done in the middleware

internal/resources/embedded/support-bundle/collect-script.sh
pmig: hmm i think that can get messy, especially for the logs, I would like to start without it and if it becomes an issue tackle it.

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,463 @@
...
+		request, err := JsonBody[api.CreateSupportBundleRequest](w, r)
+		if err != nil {
+			return
+		}
+
+		if request.Title == "" {
+			http.Error(w, "title is required", http.StatusBadRequest)
+			return
+		}
+
```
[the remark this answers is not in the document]
pmig: this is on purpose, no env --> no envs get collected

internal/handlers/support_bundles_collect.go
```diff
@@ -0,0 +1,159 @@
...
+			http.Error(w, "content file is required", http.StatusBadRequest)
+			return
+		}
+		defer file.Close()
+
+		contentBytes, err := io.ReadAll(file)
+		if err != nil {
+			http.Error(w, "failed to read content", http.StatusBadRequest)
+			return
+		}
```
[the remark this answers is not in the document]
pmig: no need to change as this is done in the middleware

internal/db/support_bundle.go
```diff
@@ -0,0 +1,401 @@
...
+	sb.collect_command,
+	sb.status_changed_by_user_account_id,
+	sb.status_changed_at,
+	u.name AS created_by_user_name,
+	u.image_id AS created_by_image_id,
+	co.name AS customer_organization_name,
+	(SELECT count(*) FROM SupportBundleResource WHERE support_bundle_id = sb.id) AS resource_count,
+	(SELECT count(*) FROM SupportBundleComment WHERE support_bundle_id = sb.id) AS comment_count,
+	(SELECT max(created_at) FROM SupportBundleComment WHERE support_bundle_id = sb.id) AS last_comment_at,
+	scu.name AS status_changed_by_user_name,
```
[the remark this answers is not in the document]
pmig: not sure about real world implication

internal/handlers/support_bundles.go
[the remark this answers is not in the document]
kosmoz: I would remove it as it currently doesn't do anything

internal/resources/embedded/support-bundle/collect-script.sh
kosmoz: I understand if we're talking about logs, but I really think a user would want to see which of their secrets are dumped to a remote server before committing...

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,476 @@
...
+			log.Error("failed to get organization", zap.Error(err))
+			sentry.GetHubFromContext(ctx).CaptureException(err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		baseURL := customdomains.AppDomainOrDefault(*org)
+		// collectCommand is stored so it can be shown again while
+		// the bundle is still in initialized state.
+		collectCommand := fmt.Sprintf(
```
kosmoz: this variable is never read before it is overwritten

internal/handlers/support_bundles.go
```diff
@@ -0,0 +1,476 @@
...
+			return
+		}
+
+		// Replace placeholder with actual bundle ID now that we have it
+		collectCommand = fmt.Sprintf(
+			"curl -fsSL '%s/api/v1/support-bundle-collect/%s/collect-script?token=%s' | sh",
+			baseURL, bundle.ID.String(), rawToken,
+		)
+		bundle.CollectCommand = &collectCommand
+		if err := db.UpdateSupportBundleCollectCommand(ctx, bundle.ID, collectCommand); err != nil {
```
kosmoz: since the collect command includes the non-hashed secret, storing the hash doesn't add any security here. In this case it would be easier to just store the unhashed secret and generate the command on-demand.

internal/handlers/support_bundles_collect.go
```diff
@@ -0,0 +1,159 @@
...
+	"go.uber.org/zap"
+)
+
+// SupportBundleScriptRouter handles endpoints called by the collect script.
+// All endpoints use query-param token auth tied to the specific bundle.
+func SupportBundleScriptRouter(r chiopenapi.Router) {
+	r.WithOptions(option.GroupTags("Support Bundles"))
+
+	r.Route("/{supportBundleId}", func(r chiopenapi.Router) {
+		r.With(auth.SupportBundleAuthentication.Middleware).Group(func(r chiopenapi.Router) {
```
kosmoz: this group is not necessary. move the middleware to the parent route

internal/handlers/support_bundles_collect.go
```diff
@@ -0,0 +1,159 @@
...
+			Content:         string(contentBytes),
+		}
+		if err := db.CreateSupportBundleResource(ctx, &resource); err != nil {
+			log.Error("failed to create support bundle resource", zap.Error(err))
+			sentry.GetHubFromContext(ctx).CaptureException(err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		RespondJSON(w, mapping.SupportBundleResourceToSummaryAPI(resource))
```
kosmoz, line 132: I think I commented on this before, but why respond with the full (potentially huge) resource here? the script doesn't use it

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,171 @@
...
+                type="button"
+                (click)="cancelBundle()"
+                [disabled]="updatingStatus()"
+                class="inline-flex items-center py-2 px-4 text-sm font-medium text-gray-700 bg-white border border-gray-300 hover:bg-gray-100 focus:ring-4 focus:outline-none focus:ring-gray-200 rounded-lg dark:text-gray-300 dark:bg-gray-800 dark:border-gray-600 dark:hover:bg-gray-700 dark:focus:ring-gray-700">
+                <fa-icon [icon]="faXmark" class="mr-2"></fa-icon>
+                Cancel
+              </button>
+            }
+            @if (auth.isVendor() && bundle.status === 'created') {
+              <button
```
kosmoz, line 34: still needs permission check

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,171 @@
...
+          class="border-t border-gray-200 dark:border-gray-700 pt-4">
+          <div class="flex gap-3">
+            <textarea
+              formControlName="content"
+              rows="3"
+              placeholder="Write a comment..."
+              class="flex-1 bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-primary-600 focus:border-primary-600 block w-full p-2.5 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-primary-500 dark:focus:border-primary-500"></textarea>
+          </div>
+          <div class="flex justify-end mt-3">
+            <button
```
kosmoz: still needs permission check

frontend/ui/src/app/support-bundles/vendor/support-bundle-settings.component.html
```diff
@@ -0,0 +1,167 @@
...
+        Back to Support Bundles
+      </a>
+      <h1
+        class="pb-4 mb-4 text-3xl font-extrabold leading-none tracking-tight text-gray-900 md:text-4xl dark:text-white border-b border-gray-200 dark:border-gray-700">
+        Support Bundle Settings
+      </h1>
+      <p class="text-gray-500 dark:text-gray-400">
+        Configure the environment variables that should be collected from Docker containers when customers create
+        support bundles. Only the variables listed below will be collected. Variables marked as redacted will have their
+        values replaced with [REDACTED] in the collected data.
```
kosmoz: Inaccurate: The variables are also collected on the host

frontend/ui/src/app/support-bundles/vendor/support-bundle-settings.component.html
```diff
@@ -0,0 +1,167 @@
...
+        @if (auth.hasAnyRole('read_write', 'admin') && envVarsArray.length > 0) {
+          <h2 class="text-xl font-bold dark:text-white mt-8 mb-2">Danger Zone</h2>
+          <div
+            class="flex flex-col rounded-lg border border-red-300 dark:border-red-800 bg-red-50 dark:bg-gray-800 divide-y divide-gray-200 dark:divide-gray-600">
+            <div class="p-4 flex items-center justify-between dark:text-white">
+              <div>
+                <div>Delete configuration</div>
+                <div class="text-xs font-normal text-gray-500 dark:text-gray-400">
+                  Permanently delete the support bundle configuration. Customers will no longer be able to create
+                  support bundles.
```
kosmoz: Inaccurate: Customers can still create support bundles

internal/resources/embedded/support-bundle/collect-script.sh
```diff
@@ -0,0 +1,168 @@
...
+  echo "---"
+  IDX=1
+  while IFS="$(printf '\t')" read -r CID CNAME CSTATUS CIMAGE; do
+    printf "  [%d] %s (%s) - %s\n" "$IDX" "$CNAME" "$CSTATUS" "$CIMAGE"
+    IDX=$((IDX + 1))
+  done <<EOF_CONTAINERS
+$CONTAINERS
+EOF_CONTAINERS
+  echo ""
+  echo "Enter container numbers to EXCLUDE (comma-separated), or press Enter to include all:"
```
kosmoz, line 135: The script needs to do more input validation here. And it should either accept whitespace between values or exit with error on invalid input. I entered `1, 2` but the second container still got collected

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,171 @@
...
+        <h2 class="text-lg font-semibold text-gray-900 dark:text-white mb-4">
+          Resources ({{ bundle.resources.length }})
+        </h2>
+        @if (bundle.resources.length === 0) {
+          <p class="text-sm text-gray-500 dark:text-gray-400">No resources collected yet.</p>
+        } @else {
+          <div class="space-y-3">
+            @for (resource of bundle.resources; track resource.id) {
+              <div class="border border-gray-200 dark:border-gray-700 rounded-lg">
+                <button
```
kosmoz, line 91: another (tiny) hover bug 

Image

frontend/ui/src/app/types/support-bundle.ts
kosmoz, line 1: why is this needed? we don't do this for any other type

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundle-list.component.ts
```diff
@@ -0,0 +1,70 @@
...
+      .getConfiguration()
+      .pipe(takeUntilDestroyed())
+      .subscribe({
+        next: () => this.configExists.set(true),
+        error: (e) => {
+          if (e instanceof HttpErrorResponse && e.status === 404) {
+            this.configExists.set(false);
+          }
+        },
+      });
```
kosmoz: The endpoint never responds with a 404 status code, so I think this doesn't do anything

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundle-list.component.ts
```diff
@@ -0,0 +1,70 @@
...
+      .pipe(takeUntilDestroyed())
+      .subscribe({
+        next: (bundles) => {
+          this.bundles.set(bundles);
+          this.loading.set(false);
+        },
+        error: () => {
+          this.loading.set(false);
+        },
+      });
```
kosmoz: you can take a look at angular resource signals, they do exactly this

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundle-list.component.ts
```diff
@@ -0,0 +1,70 @@
...
+      case 'canceled':
+        return 'bg-gray-100 text-gray-800 dark:bg-gray-900 dark:text-gray-300';
+      default:
+        return '';
+    }
+  }
+
+  protected statusLabel(status: SupportBundle['status']): string {
+    return status.charAt(0).toUpperCase() + status.slice(1);
+  }
```
kosmoz: this can be done with css https://tailwindcss.com/docs/text-transform#capitalizing-text

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundle-list.component.ts
```diff
@@ -0,0 +1,70 @@
...
+          this.bundles.set(bundles);
+          this.loading.set(false);
+        },
+        error: () => {
+          this.loading.set(false);
+        },
+      });
+  }
+
+  protected statusBadgeClass(status: SupportBundle['status']): string {
```
kosmoz:
```suggestion
  protected statusBadgeClass(status: SupportBundleStatus): string {
```

frontend/ui/src/app/support-bundles/vendor/vendor-support-bundle-list.component.ts
```diff
@@ -0,0 +1,70 @@
...
+      case 'initialized':
+        return 'bg-blue-100 text-blue-800 dark:bg-blue-900 dark:text-blue-300';
+      case 'created':
+        return 'bg-yellow-100 text-yellow-800 dark:bg-yellow-900 dark:text-yellow-300';
+      case 'resolved':
+        return 'bg-green-100 text-green-800 dark:bg-green-900 dark:text-green-300';
+      case 'canceled':
+        return 'bg-gray-100 text-gray-800 dark:bg-gray-900 dark:text-gray-300';
+      default:
+        return '';
```
kosmoz: you can use the `never` util here to make this case a compile time error

frontend/ui/src/app/support-bundles/customer/customer-support-bundles.component.html
```diff
@@ -0,0 +1,161 @@
...
+      <div class="p-4 md:p-5">
+        @if (collectCommand(); as cmd) {
+          <div class="space-y-4">
+            <div
+              class="p-4 text-sm text-blue-800 rounded-lg bg-blue-50 dark:bg-gray-800 dark:text-blue-400 border border-blue-300 dark:border-blue-800"
+              role="alert">
+              <p class="font-medium mb-2">Support bundle initialized successfully.</p>
+              <p class="mb-2">
+                The script will collect vendor defined environment variables and all docker container logs. You will
+                have the choice to select your the containers you want to collect data from. No data will be sent to the
```
kosmoz:
```suggestion
                have the choice to select the containers you want to collect data from. No data will be sent to the
```

frontend/ui/src/app/support-bundles/customer/customer-support-bundles.component.html
```diff
@@ -0,0 +1,161 @@
...
+                server, without your permission.
```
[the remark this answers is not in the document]
kosmoz: I noticed this too I think the message should be changed

internal/resources/embedded/support-bundle/collect-script.sh
kosmoz, line 1: The script does not work with stopped containers for some reason. They are shown under "available containers" but no env or logs are collected

Image

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,175 @@
...
+              <button
+                type="button"
+                (click)="cancelBundle()"
+                [disabled]="updatingStatus()"
+                class="inline-flex items-center py-2 px-4 text-sm font-medium text-gray-700 bg-white border border-gray-300 hover:bg-gray-100 focus:ring-4 focus:outline-none focus:ring-gray-200 rounded-lg dark:text-gray-300 dark:bg-gray-800 dark:border-gray-600 dark:hover:bg-gray-700 dark:focus:ring-gray-700">
+                <fa-icon [icon]="faXmark" class="mr-2"></fa-icon>
+                Cancel
+              </button>
+            }
+            @if (auth.isVendor() && bundle.status === 'created') {
```
kosmoz: should be only visible for admin and read_write

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
```diff
@@ -0,0 +1,175 @@
...
+        @if (bundle.statusChangedByUserName && (bundle.status === 'resolved' || bundle.status === 'canceled')) {
+          <p class="text-xs text-gray-500 dark:text-gray-500">
+            Set to {{ bundle.status }} by {{ bundle.statusChangedByUserName }} on
+            {{ bundle.statusChangedAt | date: 'medium' }}
+          </p>
+        }
+      </div>
+
+      <!-- Collect command card (shown for initialized bundles) -->
+      @if (bundle.collectCommand && bundle.status === 'initialized') {
```
kosmoz: Not sure it makes sense to show this to vendor and superadmin users. Also for readonly IDK…

internal/authn/supportbundle/authenticator.go
```diff
@@ -0,0 +1,49 @@
...
+			if err != nil {
+				return nil, fmt.Errorf(
+					"%w: invalid bundle ID", authn.ErrBadAuthentication,
+				)
+			}
+
+			bundle, err := db.GetSupportBundleByBundleSecret(
+				ctx, bundleID, bundleSecret,
+			)
+			if errors.Is(err, apierrors.ErrNotFound) {
```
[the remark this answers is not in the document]
pmig: we need to show this to the user, so we need to store it that way. It is only a short lived token

internal/migrations/sql/85_support_bundle_tables.up.sql
```diff
@@ -0,0 +1,51 @@
...
+CREATE TABLE SupportBundle (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    organization_id UUID NOT NULL REFERENCES Organization (id) ON DELETE CASCADE,
+    customer_organization_id UUID NOT NULL REFERENCES CustomerOrganization (id) ON DELETE CASCADE,
+    created_by_user_account_id UUID NOT NULL REFERENCES UserAccount (id),
+    title TEXT NOT NULL,
+    description TEXT,
+    status support_bundle_status NOT NULL DEFAULT 'initialized',
+    bundle_secret TEXT NOT NULL,
```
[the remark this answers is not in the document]
pmig: no

internal/resources/embedded/support-bundle/collect-script.sh
pmig: it is not possible to get env variables for stopped containers, but the information is still valuable that a container is stopped.

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
pmig: fixed

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.html
pmig: fixed

internal/resources/embedded/support-bundle/collect-script.sh
pmig: doing this now with docker inspect as a fallback

internal/resources/embedded.go
```diff
@@ -13,20 +14,21 @@ var (
...
+	}
+	tmpl, err := template.ParseFS(fsys, name)
+	if err != nil {
 		return nil, fmt.Errorf("failed to parse template %v: %w", name, err)
-	} else {
-		templates[name] = tmpl
-		return tmpl, nil
 	}
+	templates.Store(name, tmpl)
+	return tmpl, nil
```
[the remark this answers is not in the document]
kosmoz: Please no, currently we parse the template twice worst case

frontend/ui/src/app/support-bundles/dashboard-card/support-bundle-dashboard-card.component.ts
```diff
@@ -0,0 +1,34 @@
...
+      case 'resolved':
+        return 'bg-green-500';
+      case 'canceled':
+        return 'bg-gray-400';
+      default:
+        return '';
+    }
+  }
+
+  protected statusLabel(status: SupportBundle['status']): string {
```
kosmoz: See previous comment: https://github.com/distr-sh/distr/pull/1910#discussion_r2890479534

> this can be done with css https://tailwindcss.com/docs/text-transform#capitalizing-text

kosmoz: See previous comment: https://github.com/distr-sh/distr/pull/1910#discussion_r2890487873

```suggestion
  protected statusLabel(status: SupportBundleStatus): string {
```

frontend/ui/src/app/support-bundles/dashboard-card/support-bundle-dashboard-card.component.ts
```diff
@@ -0,0 +1,34 @@
...
+@Component({
+  selector: 'app-support-bundle-dashboard-card',
+  templateUrl: './support-bundle-dashboard-card.component.html',
+  imports: [RouterLink, NgClass],
+})
+export class SupportBundleDashboardCardComponent {
+  public readonly customerName = input.required<string>();
+  public readonly bundles = input.required<SupportBundle[]>();
+
+  protected statusDotClass(status: SupportBundle['status']): string {
```
kosmoz:
```suggestion
  protected statusDotClass(status: SupportBundleStatus): string {
```

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.ts
```diff
@@ -0,0 +1,206 @@
...
+      }
+      return next;
+    });
+  }
+
+  protected isResourceExpanded(resourceId: string): boolean {
+    return this.expandedResources().has(resourceId);
+  }
+
+  protected statusClass(status: SupportBundleDetail['status']): string {
```
kosmoz: I really don't like this style of type declaration

```suggestion
  protected statusClass(status: SupportBundleStatus): string {
```

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.ts
```diff
@@ -0,0 +1,206 @@
...
+      default:
+        return '';
+    }
+  }
+
+  protected shortId(id: string): string {
+    return id.substring(0, 8);
+  }
+
+  protected backRoute(): string {
```
kosmoz: isVendor doesn't change. so this could be converted to a readonly property (stuff like this will help us once we migrate to zoneless change detection)

internal/resources/embedded/support-bundle/collect-script.sh
kosmoz: I think something is broken now... I tried again with two containers (one running, one stopped). When prompted, I responded "1, 2" but no container env or logs were uploaded (not even for the running container). Checking with `docker logs ...`, both containers do have some output. Host env was uploaded without issues

Also, I think at least the logs should be uploaded for a stopped container. This should be possible and would (arguably) be pretty critical in a situatio where a customer creates a support bundle in the first place

frontend/ui/src/app/support-bundles/dashboard-card/support-bundle-dashboard-card.component.ts
pmig: resolved in a09c3a16

frontend/ui/src/app/support-bundles/detail/support-bundle-detail.component.ts
pmig: resolved in 0dc67907

PR #2031, opened 2026-03-18
chore: replace minio with rustfs

kosmoz:
Backup/restore storage data:
```shell
# create rclone config
rclone config create minio s3 provider=Other access_key_id=distr secret_access_key=distr123 endpoint=http://localhost:9000
# backup to local directory
cd (mktemp -d)
rclone sync minio:distr/ .
# restart containers with the new storage service
podman compose up -d --remove-orphans
# restore data in new storage service (bucket will be created if missing)
rclone sync . minio:distr/
```

Good to know: RustFS does not return incomplete multipart uploads if prefix is not set: https://github.com/rustfs/rustfs/issues/779

Merge https://github.com/distr-sh/distr/pull/2029 first, I will rebase this one afterwards


pmig (approved):

XavierGeerinck:
Thank you for this! I like this change, but believe it could've been a major release and include a migration path.

pmig:
> Thank you for this! I like this change, but believe it could've been a major release and include a migration path.

We highly recommend using an external object storage and database for running Distr in production and not using the bundled services.

PR #2288, opened 2026-04-16
ci(backend): use ephemeral public key if not building for tag

kosmoz:
This should allow PRs from external contributors and bots to build cleanly again (e.g. #2275)


pmig (approved):

pmig (changes requested):
actually we should ensure the migration tests still work

pmig (approved):

PR #2289, opened 2026-04-16
ci: agent changelog generation

pmig:
- change this to not be an action --> use the prebuild step for website & frontend
- investigate if we can use the changelog file


hack/agent-changelog.sh
kosmoz, line 1: I want to never touch this file, it hurts my brain to look at it.

Can we write this in a sane language? NodeJS?

hack/agent-changelog.mjs
```diff
@@ -0,0 +1,79 @@
...
+
+function parseCommit(line) {
+  const hash = line.slice(0, 40);
+  const subject = line.slice(41);
+  const m = subject.match(COMMIT_RE);
+  if (!m) return null;
+  const entry = {
+    scope: m.groups.scope,
+    description: m.groups.description,
+    commit: hash.slice(0, 8),
```
kosmoz: I think a bit better would be to get the abbreviated hash from git directly and using `indexOf(" ")` to determine where to split the string:

```suggestion
    return git(`log --first-parent --format="%h %s" ${range}`).split('\n').filter(Boolean);
  } catch {
    return [];
  }
}

function parseCommit(line) {
  const i = line.indexOf(" ");
  const hash = line.slice(0, i);
  const subject = line.slice(i + 1);
  const m = subject.match(COMMIT_RE);
  if (!m) return null;
  const entry = {
    scope: m.groups.scope,
    description: m.groups.description,
    commit: hash,
```

https://git-scm.com/docs/pretty-formats#Documentation/pretty-formats.txt-h

hack/agent-changelog.mjs
```diff
@@ -0,0 +1,79 @@
...
+  for (const line of lines) {
+    const parsed = parseCommit(line);
+    if (!parsed) continue;
+    if (!byType.has(parsed.type)) byType.set(parsed.type, []);
+    byType.get(parsed.type).push(parsed.entry);
+  }
+  if (byType.size === 0) return null;
+  return {
+    version,
+    types: [...byType.entries()].map(([type, changes]) => ({type, changes})),
```
kosmoz: I find `types` a bit confusing. Maybe change to "sections"? Release please uses "sections"

kosmoz (approved):

PR #2325, opened 2026-04-21
chore: improved agent update modal

pmig:


website/src/content/docs/privacy-policy.mdx
```diff
@@ -1,6 +1,8 @@
 ---
 title: Privacy Policy
 description: Privacy Policy
+slub: privacy-policy
```
pmig:
```suggestion
slug: privacy-policy
```

website/src/content/docs/privacy-policy.mdx
```diff
@@ -1,6 +1,8 @@
...
+hideSidebar: true
```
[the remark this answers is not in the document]
pmig: fixed

website/package.json
```diff
@@ -3,10 +3,10 @@
   "type": "module",
   "version": "0.0.1",
   "scripts": {
-    "agent-changelog": "node ../hack/agent-changelog.mjs ../CHANGELOG.md public/agent-changelog.json",
-    "predev": "pnpm agent-changelog",
-    "prestart": "pnpm agent-changelog",
-    "prebuild": "pnpm agent-changelog",
+    "copy-changelog": "cat src/content/docs/_changelog-header.md > src/content/docs/changelog.md && tail -n +2 ../CHANGELOG.md >> src/content/docs/changelog.md",
```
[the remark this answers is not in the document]
pmig: this looks way to complicated

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.html
```diff
@@ -380,6 +380,52 @@ <h3 class="font-bold truncate break-all" [title]="deploymentTarget().name">
   }
 </div>
 
+<ng-template #updateAgentConfirmTemplate>
+  <div class="space-y-4 text-sm text-gray-700 dark:text-gray-200 min-w-[42rem]">
```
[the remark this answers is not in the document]
pmig: this is needed so that the modal gets wider

website/package.json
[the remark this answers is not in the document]
kosmoz: yeah wth… we can use `shx` if that becomes a problem

frontend/ui/src/constants.ts
kosmoz, line 1: This file doesn't belong to `env`. I would move it to `src/constants/urls.ts` or `src/constants.ts`

frontend/ui/src/data/index.ts
kosmoz, line 1: `buildconfig` is no longer a good name for this package. I would rename it to `data`.

frontend/ui/src/app/services/agent-changelog.service.ts
kosmoz, line 1: These type definitions are not necessary, the Typescript compiler can infer the type from the JSON structure. After removing them, I don't think this service adds anything. Components should just import `agentChangelog` directly.

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.ts
```diff
@@ -452,6 +472,35 @@ export class DeploymentTargetCardComponent {
...
+        if (changes.length === 0) continue;
+        const merged = sections.get(section.section);
+        if (merged) {
+          merged.changes.push(...changes);
+        } else {
+          sections.set(section.section, {section: section.section, changes: [...changes]});
+        }
+      }
+    }
+    return [...sections.values()];
```
kosmoz: I know the AI LOVES to write for loops but I strongly prefer functional style pipelines when writing JS code. In this case it would be some kind of filter+map+reduce or something. 

However I don't understand why this is necessary at all? Why don't we just keep the version headings? It would make cases where there are lots of changes due to several versions skipped much easier to read

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.html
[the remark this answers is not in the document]
kosmoz: I agree with the AI here. This change would make it responsive. Also `2xl`=`[42rem]`.

.gitignore
```diff
@@ -59,4 +59,4 @@ Thumbs.db
 .serena
 
 # generated agent changelog
-frontend/ui/public/agent-changelog.json
+frontend/ui/src/buildconfig/agent-changelog.json
```
kosmoz: This is already covered by the ignore file in `buildconfig` it can be removed

frontend/ui/src/app/deployments/deployment-target-card/deployment-target-card.component.ts
pmig: reintroduced the grouped updates

kosmoz (approved):

PR #2397, opened 2026-05-07
fix: Add rollingUpdate option as null to standalone mode

ritwik-g:
With update strategy as Recreate the default rolling update options in original charts is causing conflict which is leading to helm apply failures


pmig:
@ritwik-g is this pull request resolving a specific bug? If so could you provide steps to reproduce?

ritwik-g:
> @ritwik-g is this pull request resolving a specific bug? If so could you provide steps to reproduce?

@pmig I was getting error simply with trying the helm install. The reason for error as I understood was,

- rustfs subchart has rolling update as the mode.standalone.strategy.type and the block mode.standalone.strategy.rollingUpdate is defined. https://github.com/rustfs/rustfs/blob/5f5b1207e060b08ce06c5623b3575d9035e5d848/helm/rustfs/values.yaml#L38-L42
- In distr's values.yaml we are only passing strategy.type as Recreate. So the mode.standalone.strategy.rollingUpdate is still added to the final deployment. This is getting rejected at k8s level since these are contradicting configurations.

ritwik-g:
@pmig I had tried with 2.20.1 version of distr.

Anyway looks like in the rustfs it has been [fixed](https://github.com/rustfs/rustfs/pull/2728/changes) since then and the issue might not be reproducible anymore with newer distr versions.

Still I think it might be a good idea to include this explicitly to avoid any future errors.

pmig:
@ritwik-g I think that we actually should do is improving our ci helm job to also do a helm install / upgrade against a minikube cluster so we can be sure that our helm chart is in a good state (this also helps for automated dependcy updates)

ritwik-g:
> @ritwik-g is this pull request resolving a specific bug? If so could you provide steps to reproduce?

```bash
helm template distr oci://ghcr.io/distr-sh/charts/distr \
  --version 2.20.1 --set rustfs.enabled=true \
  | awk '/charts\/rustfs\/templates\/deployment.yaml/,/^---/'
```

You can use the above code snippet to template the deployment in specific version and check the update strategy.

pmig:
thanks  @ritwik-g for your explanation I think that we will improve our ci / cd checks to ensure similar errors won't happen again.

kosmoz:
Closing, as it appears that this issue has been fixed upstream. 

https://github.com/distr-sh/distr/pull/2518 implements safeguards to prevent similar issues from happening again.

PR #2430, opened 2026-05-20
fix: enforce latin1 char set for template values

pmig:


kosmoz (commented):
I think it could also be worthwhile fixing the issue instead of surfacing the error. Maybe one of these works? https://stackoverflow.com/a/26603875

kosmoz (approved):

PR #2472, opened 2026-05-22
chore(deploy/docker): add health check to RustFS service

### LINKED ISSUE #2471: Health Check on RustFS
In `deploy/docker/docker-compose.yaml`, we don't check the health of the RustFS service.

How about adding the following:

```yaml
    restart: unless-stopped
    healthcheck:
      # Production strict TLS example (SAN/FQDN aligned, no `-k`):
      # curl -f --cacert /opt/tls/ca.crt --resolve rustfs-a.example.com:9000:127.0.0.1 https://rustfs-a.example.com:9000/health
      # curl -f --cacert /opt/tls/ca.crt --resolve rustfs-a.example.com:9001:127.0.0.1 https://rustfs-a.example.com:9001/rustfs/console/health
      test:
        [
          "CMD",
          "sh", "-c",
          "curl -f http://127.0.0.1:9000/health && curl -f http://127.0.0.1:9001/rustfs/console/health"
        ]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 40s
```

sasmaq:
closes #2471


deploy/docker/docker-compose.yaml
```diff
@@ -41,6 +41,22 @@ services:
...
+    healthcheck:
+      # Production strict TLS example (SAN/FQDN aligned, no `-k`):
+      # curl -f --cacert /opt/tls/ca.crt --resolve rustfs-a.example.com:9000:127.0.0.1 https://rustfs-a.example.com:9000/health
+      # curl -f --cacert /opt/tls/ca.crt --resolve rustfs-a.example.com:9001:127.0.0.1 https://rustfs-a.example.com:9001/rustfs/console/health
+      test:
+        [
+          'CMD',
+          'sh',
+          '-c',
+          'curl -f http://127.0.0.1:9000/health && curl -f http://127.0.0.1:9001/rustfs/console/health',
```
kosmoz: I think checking the S3 API should be enough

```suggestion
          'curl -f http://127.0.0.1:9000/health',
```

deploy/docker/docker-compose.yaml
kosmoz, line 23: Should the `hub` service depend on storage in addition to the database?

kosmoz (changes requested):

deploy/docker/docker-compose.yaml
sasmaq: Done

deploy/docker/docker-compose.yaml
sasmaq: Based on the comment in the storage services, it seems that the storage service is optional in case there is an external S3 provider.

But it is a dependency in the current structure of the Docker Compose file, so I agree with you.

What do you think about adding a comment on top to indicate that you can remove it?

```yaml
# Remove when using an external S3 provider
```

kosmoz: sounds good :+1:

sasmaq: Done

kosmoz (approved):

PR #2483, opened 2026-05-26
docs: add blog post - Container Registry Pull-Through Cache

thekubernaut:
## Summary
- add a new blog post for the Distr pull-through cache distribution flow
- include the renamed thumbnail asset under the blog assets folder
- reflect Customer Portal PAT creation and artifact/tag-level entitlements

## Verification
- pnpm format
- pnpm run build


pmig (changes requested):
- Also an additional benefit of the Distr registry is that with your images are distributed across multiple registries, you can serve all of them via Distr (different teams can continue to push to different registries)
- I also think that we should cross link and extend our registry comparison post (https://distr.sh/blog/container-image-registry-comparison/)

pmig (dismissed):

pmig (approved):

PR #2489, opened 2026-05-26
feat: role scope for personal access tokens

pmig:


internal/types/types.go
```diff
@@ -34,6 +35,28 @@ func ParseUserRole(value string) (UserRole, error) {
 	}
 }
 
+// userRoleRank orders the role hierarchy: admin > read_write > read_only.
+// Unknown roles rank above every valid role so that LessOrEqualUserRole fails
+// closed — an invalid role compares as more privileged than any real one,
+// which makes authorization checks reject it.
+func userRoleRank(r UserRole) int {
```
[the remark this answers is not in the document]
pmig: it should be safe due to the db enum

frontend/ui/src/app/components/user-role-select.component.ts
```diff
@@ -0,0 +1,64 @@
+import {Component, computed, forwardRef, input, signal} from '@angular/core';
+import {ControlValueAccessor, FormsModule, NG_VALUE_ACCESSOR} from '@angular/forms';
+import {UserRole} from '@distr-sh/distr-sdk';
+import {USER_ROLE_LABELS, userRolesAtOrBelow} from '../../util/user-role';
+
+@Component({
+  selector: 'app-user-role-select',
+  imports: [FormsModule],
```
kosmoz: Please use ReactiveForms

frontend/ui/src/app/access-tokens/access-tokens.component.html
kosmoz, line 1: I think maybe we should add the current user role wherever we say "inherit"? E.g. 

> Leave as "Inherit" to use your current role (Administrator).

(in the form)

> inherit (Administrator) 

(in the table)

internal/types/types.go
kosmoz: I agree with the AI that the behavior of this function is confusing and sounds like a bag of footguns. I think it would be fair to panic here if we always parse/validate the role beforehand

kosmoz: I would turn this into a method on `UserRole`. So it can be called as `role.Rank()` or `role.IntValue()`

internal/types/types.go
```diff
@@ -34,6 +35,28 @@ func ParseUserRole(value string) (UserRole, error) {
...
+		return 1
+	case UserRoleAdmin:
+		return 2
+	default:
+		return math.MaxInt
+	}
+}
+
+// LessOrEqualUserRole reports whether a is at most as privileged as b.
+func LessOrEqualUserRole(a, b UserRole) bool {
```
kosmoz: I think also this should be a method. Should be able to call like `role.LessThan(other)`. Also I think "less than or equal" should be written as "not greater", i.e. `!role.GreaterThan(other)`

internal/types/access_token_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	mk := func(tokenRole *UserRole, orgRole UserRole) AccessTokenWithUserAccount {
+		return AccessTokenWithUserAccount{
+			AccessToken: AccessToken{UserRole: tokenRole},
+			UserRole:    orgRole,
+		}
+	}
+
+	// Explicit PAT role caps the org role.
+	g.Expect(mk(util.PtrTo(UserRoleReadOnly), UserRoleAdmin).EffectiveUserRole()).
```
kosmoz: `util.PtrTo` should no longer be used. Use `new(UserRoleReadOnly)` instead.

frontend/ui/src/app/components/users/users.component.html
```diff
@@ -111,14 +111,10 @@
...
                 <form [formGroup]="editRoleForm" class="flex" (ngSubmit)="submitEditUserRoleForm()">
-                  <select
-                    class="bg-gray-50 border border-gray-300 text-gray-900 text-sm w-32 rounded-s-lg focus:ring-blue-500 focus:border-blue-500 block px-2 py-2 dark:bg-gray-700 dark:border-gray-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-500 dark:focus:border-blue-500"
-                    id="user-role-select"
-                    formControlName="userRole">
-                    <option value="read_only">Viewer</option>
-                    <option value="read_write">User</option>
-                    <option value="admin">Administrator</option>
-                  </select>
+                  <app-user-role-select
```
kosmoz, line 114: should we add a warning dialog if changing to a "lesser" role that PATs might become invalid if they have a greater role?

pmig: I don't see many use cases of users getting downgraded to a "lesser" role.

internal/types/types.go
pmig: done `!tok.AccessToken.UserRole.GreaterThan(tok.UserRole)`

internal/types/types.go
pmig: done

internal/types/access_token_test.go
pmig: done

internal/types/types.go
pmig: ok

internal/types/access_token_test.go
```diff
@@ -0,0 +1,36 @@
...
+			AccessToken: AccessToken{UserRole: tokenRole},
+			UserRole:    orgRole,
+		}
+	}
+
+	// Explicit PAT role caps the org role.
+	g.Expect(mk(new(UserRoleReadOnly), UserRoleAdmin).EffectiveUserRole()).
+		To(Equal(UserRoleReadOnly))
+	g.Expect(mk(new(UserRoleReadWrite), UserRoleAdmin).EffectiveUserRole()).
+		To(Equal(UserRoleReadWrite))
```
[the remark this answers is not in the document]
pmig: Verified locally — `new(UserRoleReadOnly)` does compile and the test passes (Go 1.26). This isn't the textbook `new(T)` form, but the Go compiler accepts `new(value)` and returns a pointer initialized to that value. The repo's own `internal/util/pointer.go` was deliberately changed from `return &value` to `return new(value)` in commit 075ba086 for the same reason — passing tests via `pointer_test.go` confirm it preserves the value (not the zero value).

frontend/ui/src/app/components/user-role-select.component.ts
pmig: done

frontend/ui/src/app/access-tokens/access-tokens.component.html
pmig: done

internal/authn/authinfo/token.go
```diff
@@ -24,7 +25,7 @@ func FromAuthKey(ctx context.Context, token authkey.Key) (AuthInfo, error) {
 			emailVerified:          at.UserAccount.EmailVerifiedAt != nil,
 			organizationID:         &at.OrganizationID,
 			customerOrganizationID: at.CustomerOrganizationID,
-			userRole:               &at.UserRole,
+			userRole:               util.PtrTo(at.EffectiveUserRole()),
 			rawToken:               token,
```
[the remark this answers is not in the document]
kosmoz: correct 🙂

frontend/ui/src/app/access-tokens/access-tokens.component.html
```diff
@@ -50,6 +38,13 @@
             @for (token of accessTokens$ | async; track token.id) {
               <tr class="border-t border-gray-200 dark:border-gray-600 hover:bg-gray-100 dark:hover:bg-gray-700">
                 <td class="px-4 py-3">{{ token.label }}</td>
+                <td class="px-4 py-3">
+                  @if (token.userRole; as r) {
+                    {{ r | userRoleLabel }}
+                  } @else {
+                    <span class="text-gray-400 italic">inherits ({{ currentUserRole() | userRoleLabel }})</span>
```
kosmoz:
```suggestion
                    <span class="text-gray-400 italic">{{ currentUserRole() | userRoleLabel }} (inherited)</span>
```
(personal preference)

frontend/ui/src/app/access-tokens/access-tokens.component.ts
```diff
@@ -47,9 +53,16 @@ export class AccessTokensComponent {
   private readonly overlay = inject(OverlayService);
   protected drawer: DialogRef<void> | null = null;
 
+  protected readonly currentUserRole = computed<UserRole | undefined>(() => this.auth.getClaims()?.role);
+  protected readonly inheritOptionLabel = computed(() => {
+    const role = this.currentUserRole();
+    return role ? `Inherit from my role (${USER_ROLE_LABELS[role]})` : 'Inherit from my role';
```
kosmoz:
```suggestion
    return role ? `Inherit (${USER_ROLE_LABELS[role]})` : 'Inherit from my role';
```
(personal preference; but I think it aligns better with the help text)

pmig: done

frontend/ui/src/app/access-tokens/access-tokens.component.html
pmig: done

internal/authn/authinfo/token.go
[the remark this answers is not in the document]
pmig: done

kosmoz (approved):

PR #2499, opened 2026-05-27
docs(website): unify self-hosting introduction pages

pmig:


website/src/content/docs/docs/self-hosting/index.md
kosmoz, line 1: I know this is not part of the current PR, but I don't think this entire page is very good. Particularly the first sentence makes little sense to me. 😅 

Maybe change to something like this:

> While the easiest way to use Distr is to use our [hosted offering](/get-started/), self-hosting is also an option. Distr comes as a statically compiled Go binary, packaged as a container image and has minimal dependencies:
> * A PostgreSQL database
> * An S3 compatible object storage (only if you want to use the Distr artifacts registry)
> 
> Check out our [Docker Compose](/docs/self-hosting/docker/) or [Kubernetes](/docs/self-hosting/kubernetes/) deployment options, or find out more information about the inner workings of Distr at [`github.com/distr-sh/distr`](https://github.com/distr-sh/distr/).

PR #2500, opened 2026-05-27
refactor: remove unused FormsModule imports

pmig:


frontend/ui/src/util/forms.ts
```diff
@@ -15,3 +15,18 @@ export function toggleControlsWithoutEvent(formGroup: FormGroup, enabled: boolea
...
 }
+
+/**
+ * Returns true if the form is valid. If invalid, marks every control as touched
+ * so error messages bound to `touched && invalid` become visible. Use as an
+ * early-return guard at the top of submit handlers:
+ *
+ *   if (!validate(this.form)) return;
+ */
+export function validate(form: AbstractControl): boolean {
```
kosmoz: I don't have a strong opinion about whether this is necessary, but if we do add it it might be better to use it everywhere (which would be a much bigger change tbf)

frontend/ui/src/app/deployments/deployment-modal.component.ts
```diff
@@ -88,16 +89,17 @@ export class DeploymentModalComponent {
   private readonly toast = inject(ToastService);
   private readonly deploymentTargets = inject(DeploymentTargetsService);
 
-  protected readonly deployForm = new FormControl<DeploymentFormValue | undefined>(undefined, Validators.required);
-  protected readonly loading = signal(false);
+  protected readonly form = new FormGroup({
```
kosmoz: not sure I understand why this change is necessary

kosmoz (approved):

PR #2501, opened 2026-05-27
fix: validate email verfied in middleware

pmig:


internal/authn/authinfo/db.go
```diff
@@ -36,6 +48,9 @@ func DbAuthenticator() authn.Authenticator[AuthInfo, AuthInfoWithUserAndOrganiza
 				} else if err != nil {
 					return nil, err
 				}
+				if requireEmailVerified && user.EmailVerifiedAt == nil {
+					return nil, fmt.Errorf("%w: email not verified", authn.ErrBadAuthentication)
+				}
```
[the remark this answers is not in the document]
pmig: fixed

internal/authn/authinfo/db.go
```diff
@@ -69,6 +84,8 @@ func DbAuthenticator() authn.Authenticator[AuthInfo, AuthInfoWithUserAndOrganiza
 					return nil, err
 				} else if u.UserRole != *a.CurrentUserRole() {
 					return nil, authn.ErrBadAuthentication
+				} else if requireEmailVerified && u.EmailVerifiedAt == nil {
+					return nil, fmt.Errorf("%w: email not verified", authn.ErrBadAuthentication)
```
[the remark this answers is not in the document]
pmig: fixed

internal/handlers/auth.go
```diff
@@ -46,6 +46,64 @@ func AuthRouter(r chiopenapi.Router) {
...
 	r.With(middleware.SentryUser, auth.Authentication.Middleware, middleware.RequireOrgAndRole).
 		Post("/switch-context", authSwitchContextHandler())
+	r.Route("/verify", func(r chiopenapi.Router) {
+		r.Use(middleware.SentryUser, auth.UnverifiedEmailAuthentication.Middleware)
+
+		requestVerificationMailRateLimitPerUser := httprate.Limit(
+			3,
+			10*time.Minute,
+			httprate.WithKeyFuncs(middleware.RateLimitUserIDKey),
+		)
```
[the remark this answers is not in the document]
pmig: But it works

internal/handlers/auth.go
```diff
@@ -46,6 +46,64 @@ func AuthRouter(r chiopenapi.Router) {
...
+}
+
+func authVerifyRequestHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := internalctx.GetLogger(ctx)
+	auth := auth.UnverifiedEmailAuthentication.Require(ctx)
+	userAccount := auth.CurrentUser()
+	if userAccount.EmailVerifiedAt != nil {
+		w.WriteHeader(http.StatusNoContent)
+	} else if err := mailsending.SendUserVerificationMail(ctx, *userAccount, *auth.CurrentOrg()); err != nil {
```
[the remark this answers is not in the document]
pmig: But it works without a panic, also I think it is unlikely that an organization is deleted while a verification is pending

internal/handlers/auth.go
```diff
@@ -46,6 +46,64 @@ func AuthRouter(r chiopenapi.Router) {
 	r.Post("/reset", authResetPasswordHandler)
 	r.With(middleware.SentryUser, auth.Authentication.Middleware, middleware.RequireOrgAndRole).
 		Post("/switch-context", authSwitchContextHandler())
+	r.Route("/verify", func(r chiopenapi.Router) {
```
kosmoz, line 56: Opinion: I understand why this had to be moved but I don't think `AuthRouter` is the right place for it. I think it would be better as a separate top-level route (`/api/v1/verify`)

internal/authn/authinfo/db.go
```diff
@@ -26,7 +28,19 @@ func (a DbAuthInfo) CurrentOrg() *types.Organization {
 }
 
 func DbAuthenticator() authn.Authenticator[AuthInfo, AuthInfoWithUserAndOrganization] {
+	return dbAuthenticator(true)
+}
+
+// DbAuthenticatorAllowUnverifiedEmail is like [DbAuthenticator] but does not reject
+// callers whose email_verified claim is false. Use only for endpoints involved in the
+// email verification flow itself.
+func DbAuthenticatorAllowUnverifiedEmail() authn.Authenticator[AuthInfo, AuthInfoWithUserAndOrganization] {
```
kosmoz: I don't think this is the right way to approach this. It is not an authentication concern, and this approach will return the wrong status code IMO (401 vs 403). 

I think it would be correct and actually cleaner code-wise to create a separate middleware for this

internal/handlers/auth.go
pmig: As discussed we can leave it in auth for now. verify & switch-context probably need a new home in the near future

internal/authn/authinfo/db.go
pmig: done

internal/middleware/middleware.go
```diff
@@ -220,6 +221,27 @@ var RequireOrgAndRole = auth.Authentication.ValidatorMiddleware(
...
+// auth.Authentication.Middleware so the DB-loaded user is available in the context.
+func RequireEmailVerified(handler http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		if !env.UserEmailVerificationRequired() {
+			handler.ServeHTTP(w, r)
+			return
+		}
+		ctx := r.Context()
+		if value, err := auth.Authentication.Get(ctx); err != nil {
+			http.Error(w, err.Error(), http.StatusForbidden)
```
kosmoz: not 100% sure about this but I think it should be considered a bug if this is reached and there is no AuthInfo, so a panic might be appropriate. `Authentication.Require` can be used for this

pmig: ok, done

kosmoz (dismissed):

internal/handlers/auth.go
```diff
@@ -47,8 +47,70 @@ func AuthRouter(r chiopenapi.Router) {
...
+	ctx := r.Context()
+	log := internalctx.GetLogger(ctx)
+	auth := auth.Authentication.Require(ctx)
+	userAccount := auth.CurrentUser()
+	if userAccount.EmailVerifiedAt != nil {
+		w.WriteHeader(http.StatusNoContent)
+	} else if err := mailsending.SendUserVerificationMail(ctx, *userAccount, *auth.CurrentOrg()); err != nil {
+		log.Error("failed to send verification mail", zap.Error(err))
+		w.WriteHeader(http.StatusInternalServerError)
+		sentry.GetHubFromContext(ctx).CaptureException(err)
```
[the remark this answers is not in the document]
kosmoz: I think this comment might be valid

kosmoz (commented):
see the copilot comment; I think `authVerifyRequestHandler` should probably be guarded with `RequireOrgAndRole`

kosmoz (approved):

PR #2545, opened 2026-06-02
fix(backend): change middleware order to ensure enrich sentry correctly

pmig:


internal/middleware/middleware.go
```diff
@@ -179,12 +179,11 @@ func SentryUser(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		if hub := sentry.GetHubFromContext(ctx); hub != nil {
-			if auth, err := auth.Authentication.Get(ctx); err == nil {
```
kosmoz, line 182: I agree with the review bot. please don't panic here. or alternatively add a comment that warns about this behavior

kosmoz (approved):

PR #2552, opened 2026-06-03
fix: rework user invitation / registration / verification flow

pmig:
- don't require an authenticated email for the invite flow
- directly verify an email of the users opens an invite via email (if the user gets the response via the link - he still needs to verify the mail)
- directly login a user after signup / reset / etc so the user doesn't have to login twice.
- pull useraccount auth logic into the auth package


internal/userauth/userauth.go
```diff
@@ -0,0 +1,98 @@
...
+	if err := db.CreateUserAccountOrganizationAssignment(
+		ctx, user.ID, org.ID, org.UserRole, org.CustomerOrganizationID, nil); err != nil {
+		return types.OrganizationWithUserRole{}, err
+	}
+	return org, nil
+}
+
+// GenerateLoginToken generates a default login token scoped to the user's primary organization, the same kind
+// of token that is issued on a regular login. A personal organization is created when the user has none.
+func GenerateLoginToken(ctx context.Context, user types.UserAccount) (string, error) {
```
kosmoz, line 97: the login handler should also use this, it has a lot of the same logic. probably oidc login also

internal/userauth/userauth.go
kosmoz, line 1: I'm not sure if this package is the right place for this. Or, to put it another way, I don't like that the `auth` package mixes two different concepts (authentication flow definition vs. some "user domain" methods). I think either this file should be moved (e.g. `userauth/account.go`) or authentication.go should be moved

hack/e2e-auth-test.mjs
kosmoz, line 1: should this be run in ci?

kosmoz (commented):
also don't forget to merge `main`

hack/e2e-auth-test.mjs
pmig: no as it requires email verification to be required and fetches emails from mailhog, but I still think its nice to have in the repo.

internal/userauth/userauth.go
pmig: in the OIDC its already used, the login is slightly different due to the super admin brunch, but I can refactor it

kosmoz: `authLoginOidcCallbackHandler` currently doesn't use it I think

internal/handlers/auth.go
```diff
@@ -129,6 +128,83 @@ func authVerifyConfirmHandler(w http.ResponseWriter, r *http.Request) {
...
+	body, err := JsonBody[api.AuthResetPasswordConfirmRequest](w, r)
+	if err != nil {
+		return
+	}
+	if err := body.Validate(); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	setPasswordAndLogin(w, r, body.Password, nil)
+}
```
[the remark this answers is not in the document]
pmig: password reset is also possible within the app with any authenticated jwt ...(it requires the current password though)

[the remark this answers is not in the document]
kosmoz: I do think this is a valid concern. currently we require the password or access to the users email inbox as security proof for changing the password. A hijacked JWT (or PAT) can be used to lock someone out

kosmoz (approved):

PR #2562, opened 2026-06-03
ci: add go-mailx renovate group

kosmoz:


pmig (approved):
:-)

renovate.json
```diff
@@ -95,4 +101,4 @@
       ]
     }
   ]
-}
\ No newline at end of file
+}
```
pmig, line 104: not sure why the last line changed?

pmig (changes requested):

renovate.json
kosmoz: prettier I think

pmig (approved):

PR #2574, opened 2026-06-04
docs: clarify license key revision behavior

thekubernaut:
## Summary

- Clarifies that payload and dates are immutable per token — updating them generates a new license key rather than modifying the existing one
- Notes that existing deployments are not automatically updated and the new token must be re-fetched and re-applied manually
- Removes outdated wording that said payload/dates "cannot be changed after creation" without explaining the revision mechanism
- Removes mention of customer assignment immutability from the editing section (unnecessary detail)

### DIFF
--- a/website/src/content/docs/docs/platform/licenses/license-keys.mdx
+++ b/website/src/content/docs/docs/platform/licenses/license-keys.mdx
@@ -47,9 +47,9 @@ When you request a token for a license key, Distr generates a signed JWT. Your a
 />
 
 <Aside type="note">
-  Payload, Not Before, and Expires At are **locked after creation** and cannot
-  be changed. Only the name and description can be updated later. If you need
-  different claims or dates, create a new key.
+  The name and description can be updated at any time. If you need to change the
+  payload or dates, Distr generates a new license key — the original is not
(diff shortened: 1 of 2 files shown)


website/src/content/docs/docs/platform/licenses/license-keys.mdx
```diff
@@ -130,7 +130,11 @@ Your application is responsible for deciding what to do when verification fails
 
 ## Editing a license key
 
-You can update the **name** and **description** of an existing license key. Payload, dates, and customer assignment cannot be changed after creation.
+You can update the **name** and **description** of an existing license key at any time.
+
+If you need to change the **payload**, **Not Before**, or **Expires At**, Distr generates a new license key with the updated values — the original is not modified. You can view all previously generated license keys in the **Revisions** section of the edit view.
+
+Because tokens are stateless JWTs verified offline, existing deployments are not automatically updated. To use the new license key, re-fetch the token from the portal and apply it to the customer's deployment.
```
pmig, line 137: For now this is correct, but this will change with: https://github.com/distr-sh/distr/pull/2537

pmig (approved):

PR #2647, opened 2026-06-17
fix(backend): check authkey length before type conversion

kosmoz:


internal/authkey/key.go
```diff
@@ -21,6 +21,8 @@ func Parse(encoded string) (Key, error) {
 		return Key{}, ErrInvalidAccessKey
 	} else if decoded, err := hex.DecodeString(strings.TrimPrefix(encoded, keyPrefix)); err != nil {
 		return Key{}, fmt.Errorf("%w: %w", ErrInvalidAccessKey, err)
+	} else if len(decoded) != 16 {
```
pmig, line 24: shouldn't we use the same constant used for generating the key in the first place?

pmig (changes requested):
i guess it would qualify as a magic number.

internal/authkey/key.go
kosmoz: The only other occurrence is in the `Key` type definition which cannot use a constant. so it would still be duplicated

pmig (approved):

PR #2680, opened 2026-06-23
docs(website): add an addtional onboarding step

pmig:


thekubernaut (commented):
Suggestion: remove Step 3 — "Experience"

I'd drop this whole step (role / area of expertise / company size). It's required profiling that doesn't help provision the trial or get users to value, and it partly duplicates the Job Title + Company Name we already collect in Step 1. We can enrich role/company-size later from the email domain or on the onboarding call.

Proposal: delete Step 3 and renumber the rest (5 → 4 steps). If we keep anything, only "Company Size" has real value — move it into Step 2 as a single optional dropdown.

thekubernaut (approved):

PR #2686, opened 2026-06-24
chore: super admin personal access tokens

pmig:


internal/handlers/settings_access_token.go
```diff
@@ -41,7 +43,28 @@ func createAccessTokenHandler() http.HandlerFunc {
 			return
 		}
 
-		if request.UserRole != nil {
+		if auth.IsSuperAdmin() {
+			// Super admins may only create tokens in an organization they are a
+			// member of, and those tokens are always read-only.
```
kosmoz: If I understand correctly, this is in conflict with the premise that super-admins are not supposed to be a member of any organizations

pmig: But they are created in an organization

internal/middleware/middleware.go
```diff
@@ -272,6 +272,29 @@ func RequireEmailVerified(handler http.Handler) http.Handler {
 	return http.HandlerFunc(fn)
 }
 
+// RejectAccessToken blocks requests authenticated with a personal access token (PAT). It guards
+// session- and account-management endpoints (e.g. the authenticated /auth/* routes), which are
+// meant for interactive JWT sessions only. A PAT is deliberately scoped to a single organization
+// and a capped role, so it must not be exchangeable for a different or more privileged session
+// (e.g. a super admin's read-only PAT minting a super-admin JWT via switch-context). It must run
+// after the authentication middleware.
+func RejectAccessToken(handler http.Handler) http.Handler {
```
kosmoz, line 281: Similar middlewares are called `Block...` I think this should be called `BlockAccessToken`

internal/authn/authinfo/db.go
```diff
@@ -42,21 +42,26 @@ func DbAuthenticator() authn.Authenticator[AuthInfo, AuthInfoWithUserAndOrganiza
...
-				}, nil
+				info := &SimpleAuthInfo{
+					userID:                 a.CurrentUserID(),
+					userEmail:              a.CurrentUserEmail(),
+					organizationID:         a.CurrentOrgID(),
+					customerOrganizationID: nil,
+					emailVerified:          a.CurrentUserEmailVerified(),
+					tokenScope:             a.TokenScope(),
+					userRole:               nil, // Interactive super admins don't have a role
+					isSuperAdmin:           true,
```
kosmoz, line 53: Setting this to true by default seems really scary. since we already have the user account, I think we could set it to user.isSuperAdmin right away

PR #2807, opened 2026-07-10
fix(backend): add subscription type enforcement on organization creation

kosmoz:


thekubernaut (commented):
non blocking, just minor things I found: 

1. `createOrganization` in `internal/handlers/organization.go` still pre-sets `SubscriptionType`/`Features`, which `db.CreateOrganization` now always overwrites — dead code, could be dropped.

2. The `[]types.Feature{types.FeatureLicensing}` literal duplicates `subscription.ProFeatures` (import cycle prevents reuse). Maybe add a comment linking the two spots so they don't drift if `ProFeatures` ever grows.

thekubernaut (approved):

PR #2838, opened 2026-07-15
docs(website): add zarf integration guide

pmig:


website/src/content/docs/docs/integrations/zarf.md
```diff
@@ -0,0 +1,238 @@
...
+        for pkg in zarf-package-hello-distr-*.tar.zst; do
+          zarf package publish "$pkg" oci://registry.distr.sh/your-slug
+        done
+```
+
+Notes on this job:
+
+- `needs: build` ensures the container images exist in the source registry before Zarf tries to pull them into the package.
+- `DISTR_TOKEN` is a repository secret containing your Distr [Personal Access Token](/docs/integrations/personal-access-token/).
+- The job builds one package per CPU architecture (`-a amd64` and `-a arm64`). Publishing both to the same OCI reference creates a multi-arch index, so `zarf package deploy` automatically selects the right package for the target cluster.
```
kosmoz: If I understand correctly, this can only work with tag mutability enabled. Maybe should check and document if true

website/src/content/docs/docs/integrations/zarf.md
```diff
@@ -0,0 +1,238 @@
...
+
+```shell
+zarf tools registry login registry.distr.sh -u - -p "CUSTOMER-ACCESS-TOKEN"
+zarf init --confirm
+zarf package deploy oci://registry.distr.sh/your-slug/hello-distr:1.0.0 --confirm
+```
+
+### Fully Air-Gapped Clusters
+
+For environments without any internet access, customers pull the package on a connected machine, transfer it across the air gap, and deploy from the local file:
```
kosmoz: what is a connected machine?

website/src/content/docs/docs/integrations/zarf.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Troubleshooting
+
+### Package Creation Fails with "must include a chart version"
+
+Zarf requires an explicit `version` on every chart entry, including charts referenced via `localPath`. Add `version: '###ZARF_PKG_TMPL_VERSION###'` (or the concrete chart version) to the chart definition in your `zarf.yaml`.
+
+### Deployed Pods Fail with "ImagePullBackOff"
+
+An image referenced by your Helm chart is missing from the `images` list in `zarf.yaml`. Render your chart and compare against the package definition:
```
kosmoz: this error can have a million causes
```suggestion
An image referenced by your Helm chart might me missing from the `images` list in `zarf.yaml`. Render your chart and compare against the package definition:
```

website/src/content/docs/docs/integrations/zarf.md
```diff
@@ -0,0 +1,238 @@
...
+2. A [Personal Access Token](/docs/integrations/personal-access-token/) for pushing to the registry
+3. A Helm chart for your application, with all container images published to a registry
+4. The [Zarf CLI](https://docs.zarf.dev/getting-started/install/) installed locally for testing
+
+## How the Integration Works
+
+The delivery pipeline looks like this:
+
+1. Your release workflow builds and pushes your container images and Helm chart
+2. `zarf package create` reads your `zarf.yaml`, pulls the chart and all referenced images, and writes a single `.tar.zst` archive that also includes SBOMs for every image
```
kosmoz: Here it says that zarf "pulls the chart", but below we use a `localPath` which seems to indicate that the path from the filesystem is used (i.e. not pulled)

website/astro.config.mjs
```diff
@@ -245,6 +245,15 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
                     },
                   ],
                 },
+                {
+                  label: 'Zarf',
```
kosmoz: I'm not sure about this navigation structure. Is the plan to add more zarf related articles? 

Some alternatives:

* Air-Gapped Delivery
* Packaging Tools
* Tooling Integrations

kosmoz (dismissed):

kosmoz (approved):

PR #2846, opened 2026-07-16
chore(registry): add zarf-package to ArtifactManifestType

pmig:
We are also now displaying all infered types in the list view and in the detail view


kosmoz (commented):
I have performance concerns with the artifact list query changes

PR #2910, opened 2026-07-27
feat: vulnerability management

pmig:
Open Testing:

- [x] combination of entitlements and advisories :exploding_head: :boom:


thekubernaut (commented):
When creating an advisory, the modal size changes after clicking **Add reference.** This feels a bit jumpy and makes the form harder to follow.

Long reference text can overflow the container instead of wrapping.
Image

In the Customer Portal, the advisory status is not shown. Customers therefore cannot tell whether an advisory is still active or already resolved.

The UI lets me mark a version as Fixed, but there is no way to add context about where it was fixed, such as a link to the fixed version, release notes, or security patch.

When a customer deploys the fix/security update, the advisory in the Customer Portal should reflect that this customer is no longer affected.

Streamline **fixed** and resolved **wording** --> currently inconsistent, I would go with resolved

pmig:
> When creating an advisory, the modal size changes after clicking **Add reference.** This feels a bit jumpy and makes the form harder to follow.

Fixed.

> Long reference text can overflow the container instead of wrapping.

Fixed.

> In the Customer Portal, the advisory status is not shown. Customers therefore cannot tell whether an advisory is still active or already resolved.

Fixed.

> The UI lets me mark a version as Fixed, but there is no way to add context about where it was fixed, such as a link to the fixed version, release notes, or security patch.

Do we really want to configure this for every version? This is probably something for the advisory itself?

> When a customer deploys the fix/security update, the advisory in the Customer Portal should reflect that this customer is no longer affected.

Done

> Streamline **fixed** and resolved **wording** --> currently inconsistent, I would go with resolved

Security Advisory gets resolved and a version that actually fixes a vulnerability is marked as "fixed in this version". I think these are different things? Or which exact places do you mean?


**Also there is an additional difference**:

If a vendor resolved an advisory, but the customer didn't patch his deployment I think resolved is also not correct, I guess the "big status" should still be effected?

thekubernaut (commented):
### Entitlement scoping for affected/fixed versions (customer portal) — product gap

Verified on this branch against a real customer entitlement setup.

**Expectation**
- Vendor: see the full advisory (all affected + all fixes) — OK as-is.
- Customer: only see **affected** versions they are entitled to, and only see **fixed** versions they are entitled to.
- If a fix exists but the customer is not entitled to that fix version → show **no fix** (do not disclose the version / app).
- This matters for:
  1. **Backports / version lines** — customers on 1.3 / 1.4 / 1.5 need different fixes (`1.3.1`, `1.4.1`, `1.5.1`). “Fixed in 1.5.1” is not a fix for a 1.3 customer and must not be shown if they aren’t entitled to it.
  2. **Cross-app** — a fix published only in another application the customer can’t access must not appear as their fix.

**What we verified**
Advisory had affected `Demo 1.3.0` + `Demo 1.5.0`, and fixed `Demo 1.3.1` + `Demo 1.5.1` + `hello-distr 0.4.6`. Customer org was entitled to **Demo 1.3.0 only**.

Customer portal `/security/<advisory>` still showed:
- Affected: `1.3.0` **and** `1.5.0`
- Fixed in: `1.3.1`, `1.5.1`, **and** `hello-distr 0.4.6`

So entitlements currently gate **whether the advisory is visible at all** (via an entitled affected version), but **not** which affected/fixed versions are disclosed once it’s visible.

**Ask**
Please filter customer-facing advisory version lists (API + UI) by entitlement:
- affected versions ⊆ entitled versions
- fixed versions ⊆ entitled versions
- if no entitled fixed version → render **no fix** (don’t leak other lines/apps)

internal/migrations/sql/121_advisories.up.sql
```diff
@@ -0,0 +1,107 @@
+CREATE TYPE advisory_status AS ENUM ('triage', 'draft', 'published', 'resolved', 'canceled');
+
+CREATE TYPE advisory_severity AS ENUM ('none', 'low', 'medium', 'high', 'critical');
+
+CREATE TYPE advisory_version_relation AS ENUM ('affected', 'fixed');
```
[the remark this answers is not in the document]
pmig: fixed

thekubernaut (dismissed):

frontend/ui/src/app/types/advisory.ts
kosmoz, line 1: this is not something we do

frontend/ui/src/app/advisories/advisory-form.component.html
```diff
@@ -0,0 +1,328 @@
...
+                class="distr-input"
+                placeholder="Short summary of the advisory" />
+            </div>
+
+            <div class="grid grid-cols-2 gap-4">
+              <div>
+                <label for="severity" class="distr-label mb-2">Severity</label>
+                <select formControlName="severity" id="severity" class="distr-input capitalize">
+                  @for (severity of severities; track severity) {
+                    <option [value]="severity" class="capitalize">{{ severity }}</option>
```
kosmoz, line 71: Weird: the options are lowercase, the value is capitalized. All should be the same.

frontend/ui/src/app/advisories/advisory-form.component.html
```diff
@@ -0,0 +1,328 @@
...
+              </div>
+            </div>
+
+            <div>
+              <div class="flex items-center justify-between mb-2">
+                <label for="description" class="distr-label">Description</label>
+                <button
+                  type="button"
+                  (click)="descriptionPreview.set(!descriptionPreview())"
+                  class="text-sm text-primary-600 hover:underline dark:text-primary-500">
```
kosmoz, line 93: this is a different shade of blue we use elsewhere and it's more difficult to read (at least in dark mode). Compare:

Image

With (application version form):

Image

Also, if we would use the darker dialog background that we use e.g. for the deployment, create artifact, create secret etc. dialogs it would also be easier to read.

Also, on the branding page, we use an outline button for this, shouldn't both be the same (one way or another)?

frontend/ui/src/app/advisories/advisory-form.component.html
```diff
@@ -0,0 +1,328 @@
...
+                placeholder="Add a tag and press Enter" />
+            </div>
+
+            <div>
+              <div class="flex items-center justify-between mb-2">
+                <span class="distr-label">References</span>
+                <button
+                  type="button"
+                  (click)="addReference()"
+                  class="text-sm text-primary-600 hover:underline dark:text-primary-500">
```
kosmoz, line 143: also wrong shade of blue (see previous comment)

frontend/ui/src/app/advisories/advisory-detail.component.html
```diff
@@ -0,0 +1,335 @@
...
+                  [disabled]="updatingStatus()"
+                  class="py-2 px-4 text-sm font-medium distr-btn-secondary disabled:opacity-60 disabled:cursor-not-allowed">
+                  {{ statusActionLabel(status) }}
+                </button>
+              }
+              <button
+                type="button"
+                (click)="openEditDialog(editAdvisoryDialog)"
+                class="py-2 px-4 text-sm font-medium distr-btn-secondary">
+                <fa-icon [icon]="faPen" class="me-2" />
```
kosmoz, line 66: spacing between icon and text looks wrong 

Image

frontend/ui/src/app/advisories/advisory-detail.component.html
```diff
@@ -0,0 +1,335 @@
...
+      <fa-icon [icon]="faArrowLeft" class="me-2" />
+      Back to advisories
+    </a>
+
+    @if (advisory(); as advisory) {
+      <div class="bg-white dark:bg-gray-800 shadow-md sm:rounded-lg p-4 md:p-6 mb-4">
+        <div class="flex items-start justify-between gap-4">
+          <div>
+            <h1 class="text-2xl font-semibold text-gray-900 dark:text-white">{{ advisory.title }}</h1>
+            <div class="flex flex-wrap items-center gap-2 mt-3">
```
kosmoz, line 15: This looks a bit strange (everything in a row like this)

Image

internal/types/types.go
```diff
@@ -128,16 +128,31 @@ const (
...
 // ProFeatures is the set of features granted to organizations with a paid (pro) subscription.
 var ProFeatures = []Feature{
 	FeatureLicensing,
 }
 
+// BusinessFeatures is the set granted on top of ProFeatures to the Business plan.
+var BusinessFeatures = []Feature{
+	FeaturePartnerManagement,
+	FeatureCustomDomains,
+	FeatureVulnerabilities,
```
kosmoz, line 148: Maybe I did something wrong, but shouldn't the `vulnerabilities` feature be added to orgs with an existing business subscription?

frontend/ui/src/app/advisories/advisory-detail.component.html
```diff
@@ -0,0 +1,335 @@
...
+              <ul class="space-y-2">
+                @for (reference of advisory.references; track reference.url) {
+                  <li>
+                    <a
+                      [href]="reference.url"
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      class="inline-flex items-center gap-1 text-sm text-primary-600 hover:underline dark:text-primary-500 break-all">
+                      {{ reference.label || reference.url }}
+                      <fa-icon [icon]="faUpRightFromSquare" class="text-xs" />
```
kosmoz, line 275: Minor: I think it would look better if the icon was just inline, instead of in a flex row with the text

Image

frontend/ui/src/app/advisories/advisory-detail.component.html
```diff
@@ -0,0 +1,335 @@
...
+              @for (tag of advisory.tags; track tag) {
+                <span class="distr-badge bg-gray-100 text-gray-800 dark:bg-gray-600 dark:text-gray-200">
+                  {{ tag }}
+                </span>
+              }
+            </div>
+            <p class="text-sm text-gray-500 dark:text-gray-400 mt-3">
+              Created {{ advisory.createdAt | date: 'medium' }}
+              @if (advisory.createdByUserName) {
+                by {{ advisory.createdByUserName }}
```
kosmoz, line 40: Not sure that we want to expose the vendor username. When we built the deployment timeline we explicitly decided against it

frontend/ui/src/app/advisories/advisory-detail.component.html
```diff
@@ -0,0 +1,335 @@
...
+                      <tr>
+                        @if (!auth.isCustomer()) {
+                          <th scope="col" class="p-3">Customer</th>
+                        }
+                        <th scope="col" class="p-3">Deployment Target</th>
+                        <th scope="col" class="p-3">Application</th>
+                        <th scope="col" class="p-3">State</th>
+                        <th scope="col" class="p-3">Affected Version</th>
+                        <th scope="col" class="p-3">Current Version</th>
+                        <th scope="col" class="p-3">Last Ran Affected</th>
```
kosmoz, line 141: I think this label could be misleading. If I understand correctly, it shows the last time an affected version was deployed, but I would understand it that the affected version is no longer running after this time (which may not be true!)

frontend/ui/src/app/components/activity-timeline/activity-timeline.component.html
```diff
@@ -0,0 +1,53 @@
...
+              {{ entry.body }}
+            </div>
+          }
+        </li>
+      }
+    </ol>
+  }
+
+  @if (canComment()) {
+    <form [formGroup]="commentForm" (ngSubmit)="submit()" class="mt-2">
```
kosmoz, line 40: small QOL thing: Shift+Enter to submit would be really nice

frontend/ui/src/app/advisories/advisory-list.component.ts
```diff
@@ -0,0 +1,246 @@
...
+  protected readonly statusOptions: MultiSelectOption<AdvisoryStatus>[] = advisoryStatuses.map((status) => ({
+    value: status,
+    label: statusLabel(status),
+  }));
+  protected readonly severityOptions: MultiSelectOption<AdvisorySeverity>[] = advisorySeverities.map((severity) => ({
+    value: severity,
+    label: severityLabel(severity),
+  }));
+
+  // The tags endpoint is vendor and partner only, so customers must not request it.
```
kosmoz, line 97: This seems like a strange limitation. Customers can see the tags, why not let them filter by tags?

frontend/ui/src/app/advisories/advisory-resolve-dialog.component.html
```diff
@@ -0,0 +1,56 @@
...
+        <button type="button" (click)="resolve()" class="py-2.5 px-5 text-sm font-medium distr-btn-secondary">
+          Resolve
+        </button>
+        <button
+          type="button"
+          (click)="resolveWithComment()"
+          [disabled]="!hasComment()"
+          class="text-white distr-btn-primary disabled:opacity-60 disabled:cursor-not-allowed">
+          Resolve with comment
+        </button>
```
kosmoz, line 52: why two submit actions? If I write a comment it should be with comment, if I don't it should be without

frontend/ui/src/app/advisories/advisory-list.component.html
```diff
@@ -0,0 +1,172 @@
...
+                    <td class="px-4 py-3">
+                      @if (showsAffectedState) {
+                        <span class="distr-badge" [ngClass]="affectedBadgeClass(advisory.affected)">
+                          {{ affectedLabel(advisory.affected) }}
+                        </span>
+                      } @else {
+                        <span class="distr-badge" [ngClass]="statusBadgeClass(advisory.status)">
+                          {{ statusLabel(advisory.status) }}
+                        </span>
+                      }
```
kosmoz, line 106: A customer can not see in the overview whether an advisory has been resolved, wouldn't that be relevant info? (I still don't quite understand when an advisory would be resolved)

frontend/ui/src/app/advisories/advisory-list.component.ts
```diff
@@ -0,0 +1,246 @@
...
+  protected readonly affectedLabel = affectedLabel;
+  protected readonly affectedBadgeClass = affectedBadgeClass;
+
+  /** Everyone outside the vendor organization is shown their own exposure, not the status. */
+  protected readonly showsAffectedState = this.auth.isCustomer() || this.auth.isPartner();
+
+  protected readonly routePrefix = this.auth.isCustomer() ? '/security' : '/advisories';
+  protected readonly canEdit = this.auth.isVendor() && this.auth.hasAnyRole('read_write', 'admin');
+  // Filtering by a status that the column does not show would be a dropdown of terms the
+  // reader never sees anywhere else.
```
kosmoz, line 72: This is not true, on the details page, a customer can see when an advisory was resolved.

kosmoz (commented):
* would it make sense to only add timeline entries after publishing?
* why are tag changes in a separate timeline entry and everything else is combined into one?
* after initial creation, there is a button that says "move back to triage" but it never was in triage. What does that even mean?
* Also, I'm not sure what it means to "resolve" a vulnerability? Affected users are still affected, even if I publish a fix...
* Should there be a warning when deploying an affected version? Also, any indicator on the deployment list page?

pmig:
@kosmoz please review the state machine logic

kosmoz (commented):
It looks better now but I still don't quite understand the state transition logic. My suggestion would be to just allow all transitions and change the UI to a dropdown (similar to linear), but happy to discuss further.

in addition:
* I think the CVE ID should be moved to references
* the badge spacing is inconsistent between the two rows in the header (gap-2 vs. gap-3 - I think gap-2 looks better, maybe even gap-1)
* changing tags generates a separate timeline entry, while all others are grouped

kosmoz (commented):
* Should there be an explanation that active and resolved are visible to eligible customers but the others not? maybe a warning before changing?
* small hover bug: 
  Image

* Not a blocker for now, but I think the "dropdown with badges inside" looks kind of strange. maybe we can do a little brainstorming/looking at references?

frontend/ui/src/app/advisories/advisory-list.component.html
```diff
@@ -0,0 +1,151 @@
...
+          <p class="text-sm mt-1">
+            Track a security issue from triage through disclosure and see which customers are affected.
+          </p>
+        </div>
+      }
+    }
+  </div>
+</app-page>
+
+<ng-template #createAdvisoryDialog>
```
kosmoz, line 142: this dialog is missing enter/leave animations

frontend/ui/src/app/advisories/advisory-detail.component.html
```diff
@@ -0,0 +1,359 @@
...
+          @for (tag of version.artifactVersionTags; track tag) {
+            <span [class]="tag === 'latest' ? 'distr-artifact-tag-latest' : 'distr-artifact-tag'">{{ tag }}</span>
+          }
+        </div>
+      }
+    </div>
+  </li>
+</ng-template>
+
+<ng-template #editAdvisoryDialog>
```
kosmoz, line 352: this dialog is also missing enter/leave animations

frontend/ui/src/app/advisories/advisory-display.ts
```diff
@@ -0,0 +1,207 @@
...
+): Promise<boolean> {
+  if (isCustomerVisibleStatus(from) === isCustomerVisibleStatus(to)) {
+    return true;
+  }
+  const config: ConfirmConfig = isCustomerVisibleStatus(to)
+    ? {
+        message: {
+          message: `Set this advisory to ${statusLabel(to)}?`,
+          alert: {
+            type: 'danger',
```
kosmoz: I think "danger" is a bit excessive here. It should be reserved for "very dangerous stuff" (currently the only other occurrence is deleting an organization)

frontend/ui/src/app/advisories/advisory-display.ts
```diff
@@ -0,0 +1,207 @@
...
+        confirmLabel: 'Disclose advisory',
+      }
+    : {
+        message: {
+          message: `Set this advisory to ${statusLabel(to)}?`,
+          alert: {
+            type: 'warning',
+            message:
+              'Customers and partners will no longer see this advisory. That does not retract what they ' +
+              'have already read or been notified about.',
```
kosmoz: I would remove the second part of the message, it's so obvious that it borders on being insulting

frontend/ui/src/app/advisories/advisory-form.component.html
kosmoz, line 1: should we add more padding on this dialog? it looks really crammed 🤔

kosmoz (commented):
Please resolve merge conflicts. also why a new migration?

kosmoz (approved):

PR #2935, opened 2026-08-04
chore: add turnstile registration challenge

pmig:


api/auth.go
```diff
@@ -21,6 +21,7 @@ type AuthRegistrationRequest struct {
 	OrganizationName string `json:"organizationName"`
 	Email            string `json:"email"`
 	Password         string `json:"password"`
+	TurnstileToken   string `json:"turnstileToken"`
```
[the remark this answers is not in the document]
kosmoz: I agree that it should be marked as optional, even though it is currently not exposed via the public openapi schema

frontend/ui/src/app/components/turnstile.component.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  private async render(): Promise<void> {
+    try {
+      this.turnstile = await loadTurnstile();
+      this.loadFailed.set(false);
+    } catch (e) {
+      console.error(e);
+      this.loadFailed.set(true);
+      return;
+    }
```
kosmoz: is it necessary to set `this.turnstile` in every invocation?

frontend/ui/src/app/components/turnstile.component.ts
```diff
@@ -0,0 +1,163 @@
...
+  /** Whether the challenge is unavailable, meaning no token can be obtained until the page is loaded again. */
+  public readonly unavailable = this.loadFailed.asReadonly();
+
+  private readonly colorScheme = inject(ColorSchemeService).colorScheme;
+  private readonly theme = computed<TurnstileTheme>(() => (this.colorScheme() === 'dark' ? 'dark' : 'light'));
+
+  private readonly widget = viewChild.required<ElementRef<HTMLElement>>('widget');
+  private turnstile?: TurnstileApi;
+  private widgetId?: string;
+  private renderedTheme?: TurnstileTheme;
```
kosmoz: why do we need to track this? the signal should only emit a value if it changes

frontend/ui/src/app/components/turnstile.component.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+}
+
+/**
+ * Renders the Cloudflare Turnstile widget. The token it emits has to be sent along with the form and is verified
+ * by the backend, which is the only place it means anything.
+ */
+@Component({
+  selector: 'app-turnstile',
+  changeDetection: ChangeDetectionStrategy.Eager,
```
kosmoz: why does it need eager change detection? shouldn't it only depend on the signal values?

frontend/ui/src/app/register/register.component.html
```diff
@@ -88,15 +88,18 @@ <h1 class="text-xl font-bold leading-tight tracking-tight text-gray-900 md:text-
...
           }
           <button
             type="submit"
-            [disabled]="loading"
+            [disabled]="loading()"
             class="w-full text-white bg-primary-600 hover:bg-primary-700 focus:ring-4 focus:outline-none focus:ring-primary-300 font-medium rounded-lg text-sm px-5 py-2.5 text-center dark:bg-primary-600 dark:hover:bg-primary-700 dark:focus:ring-primary-800">
             Create an account
           </button>
+          @if (turnstileSiteKey(); as siteKey) {
+            <app-turnstile [siteKey]="siteKey" (token)="onTurnstileToken($event)" />
```
kosmoz: since the challenge UI is invisible (at least for me during testing), this leads to double the spacing between the "Create an account" button and "Already have an account?" label

Image

api/auth.go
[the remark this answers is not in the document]
pmig: done

frontend/ui/src/app/components/turnstile.component.ts
```diff
@@ -0,0 +1,142 @@
...
+        resolve(loaded);
+      } else {
+        reject(new Error('the Turnstile script did not install its API'));
+      }
+    });
+    script.addEventListener('error', () => reject(new Error('failed to load the Turnstile script')));
+    document.head.appendChild(script);
+  });
+  return api;
+}
```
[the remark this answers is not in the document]
pmig: I don't see that edge case

frontend/ui/src/app/components/turnstile.component.ts
```diff
@@ -0,0 +1,163 @@
...
+    afterNextRender(() => this.render());
+    // The theme is fixed when a widget is rendered, so switching the color scheme means rendering it again. That
+    // discards the current token: challenges are solved per widget, and this is a new one.
+    effect(() => {
+      const theme = this.theme();
+      if (this.widgetId !== undefined && theme !== this.renderedTheme) {
+        this.remove();
+        this.token.emit(undefined);
+        void this.render();
+      }
```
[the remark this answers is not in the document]
pmig: fixed

frontend/ui/src/app/components/turnstile.component.ts
pmig: fixed

frontend/ui/src/app/register/register.component.html
pmig: implemented it in away that the captcha sneaks in the flow (negative top margin and bottom margin)

internal/handlers/auth.go
```diff
@@ -451,6 +454,31 @@ func authRegisterHandler(w http.ResponseWriter, r *http.Request) {
...
+	// Resolution is best-effort in the same way as in the portal endpoint: a failed lookup leaves the default
+	// host, which is the one that requires a challenge.
+	host, err := resolvePortalHost(ctx, validation.NormalizeHostname(r.Host))
+	if err != nil {
+		log.Warn("failed to resolve host for challenge verification", zap.Error(err))
+		sentry.GetHubFromContext(ctx).CaptureException(err)
+	}
+	if host.turnstileSiteKey() == nil {
+		return true
+	}
```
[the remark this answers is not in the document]
pmig: on cname domains registration should not be enabled anyways

frontend/ui/src/app/components/turnstile.component.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  private async render(theme: TurnstileTheme): Promise<void> {
+    let turnstile: TurnstileApi;
+    try {
+      turnstile = await loadTurnstile();
+    } catch (e) {
+      console.error(e);
+      this.loadFailed.set(true);
+      return;
+    }
```
kosmoz: I was thinking more something like 
```ts
if (!this.turnstile) {
  // imagine try..catch here
  this.turnstile = await loadTurnstile()
}
this.turnstile.render(/* ...*/)
```
as that would make the render function run in one go instead of making a round trip through the event loop. but it's probably not worth doing another round just because of this, I just wanted to say it fyi 😄

kosmoz (dismissed):

frontend/ui/src/app/components/turnstile.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    this.widgetId = turnstile.render(this.widget().nativeElement, {
+      sitekey: this.siteKey(),
+      action,
+      theme,
+      // Only shows the widget when the challenge actually has to be solved by hand.
+      appearance: 'interaction-only',
+      callback: (token) => this.token.emit(token),
+      'error-callback': () => this.token.emit(undefined),
+      'expired-callback': () => this.token.emit(undefined),
+    });
```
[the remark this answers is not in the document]
pmig: there is no theme switch on the registration page

PR #3036, opened 2026-08-21
chore: allow overwriting the docker socket endpoint

pmig:


kosmoz (commented):
it works but the naming is inconsistent: sometimes we use "docker endpoint" and other times "docker socket path". but I think it should be the same?

pmig:
> it works but the naming is inconsistent: sometimes we use "docker endpoint" and other times "docker socket path". but I think it should be the same?

docker endpoint always refferes to the full endpoint `unix:///....`  where dockerSocketPath begins with `/a/b/c`.

kosmoz:
> > it works but the naming is inconsistent: sometimes we use "docker endpoint" and other times "docker socket path". but I think it should be the same?
> 
> docker endpoint always refferes to the full endpoint `unix:///....` where dockerSocketPath begins with `/a/b/c`.

got it, that wasn't clear to me

kosmoz (approved):

PR #3037, opened 2026-08-21
ci: disable hostnetwork for docker builds

pmig:


kosmoz (commented):
For me it doesn't work without this flag, that's why it was added in the first place...

pmig:
ok without buildkit it actually also works with the host network

PR #3059, opened 2026-08-24
chore: improve oidc settings page, make slug unique on domain, not org

pmig:


internal/migrations/sql/124_custom_oidc_per_domain_unique.down.sql
```diff
@@ -0,0 +1,15 @@
+-- the organization-wide constraints cannot be restored while a name or slug is used on more than one
+-- domain, so the newer of every conflicting pair goes away, together with the identities linked to it
+DELETE FROM CustomOIDCConfiguration c
+  WHERE EXISTS (
+    SELECT 1 FROM CustomOIDCConfiguration other
+    WHERE other.organization_id = c.organization_id
+      AND (other.name = c.name OR other.slug = c.slug)
+      AND (other.created_at, other.id) < (c.created_at, c.id)
+  );
```
kosmoz: super edge case I guess, but I'd rather have the migration fail than deleting some random configs

frontend/ui/src/app/organization-settings/custom-oidc.component.html
```diff
@@ -131,8 +131,16 @@ <h4 class="text-base font-semibold text-gray-900 dark:text-white mt-8">Identity
       @if (!customerScoped()) {
         <h3 class="text-lg font-bold dark:text-white">Customer Portal</h3>
         <p class="text-sm font-normal text-gray-500 dark:text-gray-400">
-          One domain for all of your customers. Providers configured here apply to every customer, including ones you
-          create later. A customer that needs its own provider gets the Identity Provider feature instead.
+          One domain default for all of your customers. A provider is only ever offered on the domain it is attached to,
+          so the customer portal domain below has to exist before you can configure one here; once it does, the
+          providers you add apply to every customer, including ones you create later.
+        </p>
```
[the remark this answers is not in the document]
kosmoz: I agree, "domain default" sounds strange and I'm not sure what it means

frontend/ui/src/app/organization-settings/custom-oidc.component.html
```diff
@@ -131,8 +131,16 @@ <h4 class="text-base font-semibold text-gray-900 dark:text-white mt-8">Identity
...
+        <p class="text-sm font-normal text-gray-500 dark:text-gray-400">
+          A single customer can override all of this with a domain and providers of its own, configured on that
```
kosmoz: > A single customer ...

Can't all customers do that? 🙂 

Also, "its" sounds a bit strange to me... 

how about something like this:

> Customers can override all of this with their own domain and providers, configured on their own settings page

kosmoz (commented):
there is a bunch of small wording issues noted in the umbrella ticket, would it make sense to also address those in this PR?

internal/migrations/sql/124_custom_oidc_per_domain_unique.down.sql
pmig: done

frontend/ui/src/app/organization-settings/custom-oidc.component.html
pmig: done

PR #3065, opened 2026-08-24
build: add website to format task

pmig:
Agents often forget to format the website after a docs update. Also this commit shouldn't effect GitHub Actions.


kosmoz (commented):
If we're already touching on this, can we maybe also tell the agents that it's unnecessary to run both formatting AND linting? the cursor agent always does that....

kosmoz (approved):

PR #3111, opened 2026-08-31
refactor(frontend): consistent text input and select styling

kosmoz:


pmig (changes requested):
This is definitely the pr to extract this tailwind class soup into a single shared `distr-input` variant class using  @ apply style.css (if different from the existing one)

pmig (approved):

PR #3114, opened 2026-08-31
chore(backend): add deleting orphaned artifact versions

### LINKED ISSUE #2932: Artifact blob cleanup can never reclaim: untagged (digest-named) ArtifactVersions are never deleted
### Version

Distr 2.23.0 (community), self-hosted, using the built-in OCI registry with S3-compatible storage.

### Summary

`ArtifactBlob` cleanup can never reclaim blobs for images pushed by a normal OCI client, because **every push creates two `ArtifactVersion` rows and only one of them is ever deletable**.

An OCI push does two manifest PUTs: one by digest (`PUT /v2/<repo>/manifests/sha256:…`) and one by tag. Distr records an `ArtifactVersion` for each, so a tagged image ends up with a tag-named row *and* a digest-named row.

`db.GetAllReferencedBlobDigests` treats a blob as referenced if **any** `ArtifactVersion` mentions it:

```sql
SELECT manifest_blob_digest FROM ArtifactVersion
UNION ALL
SELECT artifact_blob_digest FROM ArtifactVersionPart
```

But nothing deletes the digest-named row:

- `deleteArtifactTagHandler` → `db.DeleteArtifactVersion(ctx, artifactID, tagName)` removes only the row named by that tag, leaving the sibling digest-named row (and its `ArtifactVersionPart` rows) intact.
- There is no per-version delete endpoint — the artifact router exposes only `DELETE /artifacts/{artifactId}` and `DELETE /artifacts/{artifactId}/tags/{tagName}`.
- The UI/API only lists *tagged* versions, so these rows are invisible to operators.

Net effect: `artifact-blob-cleanup` runs, finds every blob still referenced, and deletes nothing — indefinitely.

### Observed

On a registry used for ephemeral preview environments (frequent pushes, short-lived tags):

| | |
|---|---|
| digest-named `ArtifactVersion` rows | 9,128 |
| tag-named rows | 1,743 |
| digest-named rows with no tag sharing their digest | 7,535 |
| blobs referenced *only* by those rows | 12,401 (~80 GB) |

Cleanup job output while ~80 GB was unreclaimable:

```
ArtifactBlob cleanup finished  skippedMinAge=18  skippedStillReferenced=16402  deleted=0
```

Deleting tags (via `crane delete <repo>:<tag>`, i.e. the registry API) did **not** help: `stillReferenced` stayed flat, because the digest-named rows survive. After deleting the orphaned rows directly in Postgres, the very next cleanup run reported `deleted=12370, errors=0` and the storage was reclaimed — which confirms the bookkeeping, not the S3 deletion, is the blocker.

### Steps to reproduce

1. Push a tagged image to the Distr registry with any standard client (`docker push`, `crane copy`, `helm push`).
2. `SELECT name FROM artifactversion WHERE artifact_id = …` → two rows, one `sha256:…`, one the tag.
3. Delete the tag (API or registry `DELETE`).
4. The `sha256:…` row remains, and `artifact-blob-cleanup` continues to report those blobs as `stillReferenced` forever.

### Also worth noting

`DELETE /v2/<repo>/manifests/sha256:…` returns `UNSUPPORTED: the request is invalid or not supported`, so operators can't clear the canonical manifest through the registry API either. Deleting rows in the database is currently the only route, which is not something most operators should be doing.

### Possible fixes

1. When deleting a tag, also delete the digest-named version if no other tag references that digest (probably the smallest change, and matches what most registries do — deleting the last tag makes the manifest garbage).
2. A cleanup job for untagged versions, mirroring `ArtifactBlob` cleanup with its own min-age setting, for the backlog that already exists.
3. Expose per-version deletion in the API so untagged versions are at least addressable, and surface them in the UI.

Happy to test a patch — this is reproducible on demand in our environment.

kosmoz:


frontend/ui/src/app/artifacts/artifact-versions/artifact-versions.component.ts
```diff
@@ -327,7 +327,7 @@ export class ArtifactVersionsComponent {
   public deleteArtifactTag(artifact: ArtifactWithTags, version: TaggedArtifactVersion, tagName: string): void {
     this.overlay
       .confirm(
-        `This will untag "${tagName}" from ${artifact.name}. The artifact version SHA (${version.digest.substring(0, 12)}) will remain in the database. Are you sure?`
+        `This will untag "${tagName}" from ${artifact.name}. The artifact version (${version.digest.substring(0, 12)}) is deleted as well, unless another tag or a multi-arch index still references it. Are you sure?`
```
pmig: This message should be smarter as we can _assume_ what will happen to the sha:

like:
 - The image will only be untagged the sha will still count against your quot as another tag references it.
 - This image will be deleted as no other tag still references it:
 
 
 
image

internal/db/artifacts.go
```diff
@@ -1323,6 +1316,155 @@ func DeleteArtifactVersion(ctx context.Context, artifactID uuid.UUID, tagName st
...
+// LockArtifactExclusive locks an artifact against a concurrent deletion of one of its tags, which
+// deletes a different row and would therefore not conflict, but whose uncommitted tag each deletion
+// still reads as a reason to keep a manifest. It cannot cover a concurrent push, whose manifests are
+// unreferenced across several requests with no transaction spanning them.
+func LockArtifactExclusive(ctx context.Context, artifactID uuid.UUID) error {
+	db := internalctx.GetDb(ctx)
+	// Advisory locks share one namespace instance-wide, hence the entity prefix in the key.
+	if _, err := db.Exec(
+		ctx,
+		`SELECT pg_advisory_xact_lock(hashtextextended('Artifact:' || @artifactId::TEXT, 0))`,
```
pmig, line 1460: Postgres has native locking? Interesting!

pmig (changes requested):

frontend/ui/src/app/artifacts/artifact-versions/artifact-versions.component.ts
kosmoz: fixed

pmig (approved):

PR #3128, opened 2026-09-02
chore(frontend): introduce org filter and badges

pmig:


frontend/ui/src/app/components/search-bar.component.ts
```diff
@@ -29,8 +30,19 @@ let nextId = 0;
 export class SearchBarComponent {
   public readonly control = input.required<FormControl>();
   public readonly placeholder = input.required<string>();
+  public readonly autofocus = input(false);
```
kosmoz, line 34: we have an autofocus directive, I think it would be better to extend it with a boolean input instead of reimplementing

```html
<input [autofocus]="foo()">
```

frontend/ui/src/app/components/subscription-badge.component.scss
```diff
@@ -0,0 +1,60 @@
...
+  @apply inline-flex items-center px-2 py-0.5 rounded-md text-xs font-medium capitalize;
+  @apply bg-gray-100 text-gray-800;
+  border: 1px solid color-mix(in oklab, currentcolor 30%, transparent);
+
+  @include dark {
+    @apply bg-transparent text-gray-300;
+  }
+}
+
+:host([data-plan='pro']) {
```
kosmoz: please use CSS classes instead

frontend/ui/src/app/components/nav-bar/nav-bar.component.ts
```diff
@@ -141,8 +182,46 @@ export class NavBarComponent implements OnInit {
...
+  protected switchToHighlightedOrg(event: Event) {
+    const org = this.matchingOrgs()[this.highlightedOrgIndex()];
+    const isOptionFocused = this.orgOptions().some((option) => option.nativeElement === event.target);
+    if (!this.organizationsOpened() || !org || isOptionFocused) {
+      return;
+    }
+    // the trigger button keeps the focus after a click, and would toggle the dropdown on Enter
+    event.preventDefault();
+    this.switchContext(org);
+  }
```
kosmoz: IMO this is unnecessary, the browsers keyboard navigation works perfectly fine?

frontend/ui/src/app/components/subscription-badge.component.scss
pmig: done

frontend/ui/src/app/components/nav-bar/nav-bar.component.ts
pmig: > IMO this is unnecessary, the browsers keyboard navigation works perfectly fine?

Browser navigation only works with tab, but i guess it is ok!

kosmoz (approved):

PR #3188, opened 2026-09-09
chore: introduce generic input trimming

pmig:
- i want to introduce introduce an inceptor for the frontend
- i want to introduce introduce a generic mapper for the backend
- i want to remove the existing trimming which would be redundant


kosmoz (commented):
Please do not trim:
- everything markdown
- support bundle resource
- agent logs
- deployment logs

kosmoz (approved):

PR #3191, opened 2026-09-09
feat!: add access token key secrets

pmig:


internal/db/access_token.go
```diff
@@ -121,6 +221,94 @@ func GetAccessTokenByKeyUpdatingLastUsed(
...
+func MarkAccessTokenUsed(ctx context.Context, id uuid.UUID, slot *types.AccessTokenSecretSlot) error {
+	db := internalctx.GetDb(ctx)
+	secretExpr := ""
+	if slot != nil {
+		secretExpr = fmt.Sprintf(", %v_last_used_at = now()", accessTokenSecretPrefix(*slot))
+	}
+	if _, err := db.Exec(
+		ctx,
+		fmt.Sprintf("UPDATE AccessToken SET last_used_at = now()%v WHERE id = @id", secretExpr),
+		pgx.NamedArgs{"id": id},
```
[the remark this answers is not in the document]
pmig: fixed

internal/migrations/sql/132_access_token_secrets.up.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE AccessToken
+  ADD COLUMN secret_1_salt BYTEA,
```
kosmoz, line 2: As discussed, I think salting doesn't buy us anything here

internal/mapping/access_token.go
```diff
@@ -8,10 +8,26 @@ import (
 func AccessTokenToAPI(model types.AccessToken) api.AccessToken {
 	return api.AccessToken{
 		ID:         model.ID,
+		KeyID:      model.KeyID(),
```
kosmoz, line 11: This returns full legacy PATs which breaks our previous promise that they can not be seen a second time. I would only set the key ID for PATs with at least one secret or at least use a truncated version for ones without.

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,227 @@
...
+        @if (legacy()) {
+          <p class="mt-4 max-w-3xl text-sm text-gray-500 dark:text-gray-400">
+            This token has no secret, so it is stored as it is and shown here in full.
+          </p>
+        } @else {
+          <p class="mt-4 max-w-3xl text-sm text-gray-500 dark:text-gray-400">
+            A token has the format
+            <code class="text-gray-900 dark:text-white">distr-&lt;key id&gt;_&lt;secret&gt;</code>
+            where the secret is stored hashed and only visible on creation.
+          </p>
```
kosmoz, line 84: Not sure this information is relevant to a user

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,227 @@
...
+                Key ID
+              }
+            </dt>
+            <dd class="flex items-center gap-2">
+              <code
+                data-ph-mask-text="true"
+                class="select-all px-1 py-0.5 bg-gray-100 dark:bg-gray-700 rounded text-sm text-gray-900 dark:text-white border border-gray-300 dark:border-gray-600">
+                {{ token.keyId }}
+              </code>
+              <app-clip [clip]="token.keyId" />
```
kosmoz, line 102: I think displaying it like this is confusing. It suggests that this is the complete usable token, which it isn't. I think it would be clearer to indicate that this is a truncated token, e.g. `distr-1s4UIgDiGZf5JXKdjL2e5y…` and removing the app-clip

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,227 @@
...
+          (save)="saveLabel($event)" />
+
+        @if (legacy()) {
+          <p class="mt-4 max-w-3xl text-sm text-gray-500 dark:text-gray-400">
+            This token has no secret, so it is stored as it is and shown here in full.
+          </p>
+        } @else {
+          <p class="mt-4 max-w-3xl text-sm text-gray-500 dark:text-gray-400">
+            A token has the format
+            <code class="text-gray-900 dark:text-white">distr-&lt;key id&gt;_&lt;secret&gt;</code>
```
kosmoz, line 82:
```suggestion
            <code class="text-gray-900 dark:text-white">distr-&lt;key_id&gt;_&lt;secret&gt;</code>
```
to me it looks weird with a space

frontend/ui/src/app/access-tokens/access-token-detail.component.ts
```diff
@@ -0,0 +1,218 @@
...
+  protected readonly inheritOptionLabel = computed(() => {
+    const role = this.currentUserRole();
+    return role ? `Inherit (${USER_ROLE_LABELS[role]})` : 'Inherit from my role';
+  });
+  protected readonly createdToken = signal<AccessTokenWithKey | null>(null);
+  protected readonly savingLabel = signal(false);
+
+  protected readonly settingsForm = new FormGroup({
+    userRole: new FormControl<UserRole | undefined>(undefined),
+    expiresAt: new FormControl('', {nonNullable: true}),
```
kosmoz, line 100: Being able to change the expiration date of a token seems risky. Other platforms don't allow that. On GitHub, you can re-generate an expired token with a new expiration date, which generates a new token. maybe it would make sense to have the expiration date on the token secret instead, so one could add a new secret with a different expiration, then it would not be necessary to be able to change it

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,227 @@
+<app-page>
```
kosmoz, line 1: This page might look better as a "narrow" page

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,227 @@
...
+
+        @if (legacy()) {
+          <div class="mx-4 mb-4 flex items-start gap-2 text-sm text-gray-500 dark:text-gray-400">
+            <fa-icon [icon]="faTriangleExclamation" class="mt-0.5 text-yellow-500" />
+            <span>
+              This token was created before tokens were split into a key and a secret, so the token itself is the whole
+              credential. Securing it gives it a secret and replaces the token, so anything still using the old one has
+              to be updated.
+            </span>
+          </div>
```
kosmoz, line 177: I think this should use the common "alert" styling

frontend/ui/src/app/access-tokens/access-token-detail.component.ts
kosmoz, line 100: When I change the date input to an invalid value, it immediately sends a PATCH request that disables expiration entirely

frontend/ui/src/app/access-tokens/access-token-detail.component.ts
```diff
@@ -0,0 +1,218 @@
...
+        // Reload either way, so that a rejected change is replaced by what the server still has
+        // instead of staying on screen.
+        this.accessTokens.reload();
+      })
+    );
+  }
+
+  public async createSecret() {
+    const confirmation = this.legacy()
+      ? `Token '${this.name()}' is still stored in plain text. Securing it replaces it with a new token, ` +
```
kosmoz, line 180: This is technically wrong and makes it sound worse than it is. The token is encrypted but unhashed

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,227 @@
...
+            [tokenKey]="created.key"
+            [dismissible]="true"
+            (dismissed)="createdToken.set(null)" />
+        }
+
+        @if (legacy()) {
+          <div class="mx-4 mb-4 flex items-start gap-2 text-sm text-gray-500 dark:text-gray-400">
+            <fa-icon [icon]="faTriangleExclamation" class="mt-0.5 text-yellow-500" />
+            <span>
+              This token was created before tokens were split into a key and a secret, so the token itself is the whole
```
kosmoz, line 173: > the token itself

I think this wording is confusing. To a user "the token itself" is always the usable unit. Maybe we should update the wording to use "key" for the update-able unit and "token" for the usable credential?

internal/handlers/settings_access_token.go
kosmoz, line 1: Not a regression but it is more visible now: Do we want to allow creating tokens with an expiration date in the past? Also, does it make sense to allow setting a past expiration date on an existing token?

internal/handlers/settings_access_token.go
```diff
@@ -94,3 +152,153 @@ func deleteAccessTokenHandler() http.HandlerFunc {
...
+		return false
+	}
+	effective := membershipRole
+	if requested != nil {
+		effective = *requested
+	}
+	return !effective.GreaterThan(*callerRole)
+}
+
+func newAccessTokenSecret(key authkey.Secret) (types.AccessTokenSecret, error) {
```
kosmoz, line 298: IMO this should be moved to internal/types

frontend/ui/src/app/access-tokens/access-token-detail.component.ts
pmig: fixed, the expired component now disallows invalid and past dates

internal/handlers/settings_access_token.go
pmig: no fixed!

frontend/ui/src/app/access-tokens/access-token-detail.component.ts
```diff
@@ -0,0 +1,218 @@
...
+        // value it was started with back, and unsubscribing does not undo one the server has
+        // already accepted.
+        concatMap(({userRole, expiresAt}) =>
+          this.patch({
+            userRole: userRole ?? null,
+            // The picker works in local dates, and new Date() would read one as UTC midnight,
+            // which moves the day for everyone west of it.
+            expiresAt: expiresAt ? dayjs(expiresAt).toDate() : null,
+          })
+        ),
```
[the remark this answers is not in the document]
pmig: fixed

internal/handlers/settings_access_token.go
pmig: fixed

internal/handlers/settings_access_token.go
pmig: done

internal/migrations/sql/132_access_token_secrets.up.sql
pmig: removed

frontend/ui/src/app/access-tokens/access-token-detail.component.ts
pmig: fixed

pmig: fixed

frontend/ui/src/app/access-tokens/access-token-detail.component.html
pmig: simplified

frontend/ui/src/app/access-tokens/access-token-detail.component.ts
pmig: removed

frontend/ui/src/app/access-tokens/access-token-detail.component.html
pmig: moved to docs

api/access_token.go
```diff
@@ -3,31 +3,76 @@ package api
...
 type AccessTokenWithKey struct {
 	AccessToken
-	Key authkey.Key `json:"key"`
+	Key string `json:"key"`
 }
 
 type CreateAccessTokenRequest struct {
+	// ExpiresAt is the expiration of the secret the token is created with, and a token without one
+	// never expires. It cannot be changed afterwards, so a token is kept alive by adding a secret
+	// that expires later.
```
kosmoz:
```suggestion
```

api/access_token.go
```diff
@@ -3,31 +3,76 @@ package api
...
-	UserRole   *types.UserRole `json:"userRole,omitempty"`
+	ID uuid.UUID `json:"id"`
+	// KeyID is the beginning of the token, so that a user can tell which of their tokens a client
+	// is configured with. It stays the same when the secrets are rotated, and covers only half of
+	// the key, since a token is shown in full exactly once.
+	KeyID     string    `json:"keyId"`
+	CreatedAt time.Time `json:"createdAt"`
+	// ExpiresAt is when the token stops working, which is the last of its secrets to expire, since
+	// every secret that is still valid authenticates it. It is not settable: an expiration belongs
+	// to the secret it was created with.
```
kosmoz:
```suggestion
```

api/access_token.go
```diff
@@ -3,31 +3,76 @@ package api
...
-	ID         uuid.UUID       `json:"id"`
-	CreatedAt  time.Time       `json:"createdAt"`
-	ExpiresAt  *time.Time      `json:"expiresAt,omitempty"`
-	LastUsedAt *time.Time      `json:"lastUsedAt,omitempty"`
-	Label      *string         `json:"label,omitempty"`
-	UserRole   *types.UserRole `json:"userRole,omitempty"`
+	ID uuid.UUID `json:"id"`
+	// KeyID is the beginning of the token, so that a user can tell which of their tokens a client
+	// is configured with. It stays the same when the secrets are rotated, and covers only half of
+	// the key, since a token is shown in full exactly once.
```
kosmoz:
```suggestion
```

api/access_token.go
```diff
@@ -3,31 +3,76 @@ package api
...
+	KeyID     string    `json:"keyId"`
+	CreatedAt time.Time `json:"createdAt"`
+	// ExpiresAt is when the token stops working, which is the last of its secrets to expire, since
+	// every secret that is still valid authenticates it. It is not settable: an expiration belongs
+	// to the secret it was created with.
+	ExpiresAt  *time.Time          `json:"expiresAt,omitempty"`
```
kosmoz, line 17: I'm not sure if this field still makes sense for a token with multiple secrets. Taking the latest secret expiration seems kind of random too

api/access_token.go
```diff
@@ -3,31 +3,76 @@ package api
...
+
+func (r CreateAccessTokenRequest) Validate() error {
+	return validateAccessTokenExpiresAt(r.ExpiresAt)
+}
+
+// PatchAccessTokenRequest carries the only part of a token that is not fixed when it is created.
+// An omitted label is left unchanged and an explicit null clears it.
+type PatchAccessTokenRequest struct {
+	Label Nullable[string] `json:"label"`
+}
```
kosmoz: with the label being the only remaining field, this is strange. why would I send an empty PATCH request to do nothing? I think we can remove the Nullable here (which is kind of a hack anyways) and use nil to clear the label. also, should an empty string not also clear the label?

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,241 @@
...
+
+        @if (legacy()) {
+          <div
+            class="mx-4 mb-4 flex items-start gap-3 p-4 text-sm text-yellow-800 rounded-lg bg-yellow-50 dark:bg-transparent dark:text-yellow-300 dark:border dark:border-yellow-300"
+            role="alert">
+            <fa-icon [icon]="faTriangleExclamation" class="mt-0.5" />
+            <span>
+              Please migrate this access token to the new, more secure format that adds a secret to it. Adding one
+              issues a new token and invalidates the current one, so anything still using the current one has to be
+              updated.
```
kosmoz:
```suggestion
              Please migrate this access token to the new, more secure format. Adding a secret
              will issue a new token and invalidate the current one, so anything still using the current one has to be
              updated.
```

frontend/ui/src/app/access-tokens/access-token-detail.component.html
```diff
@@ -0,0 +1,241 @@
...
+          size="lg"
+          [value]="token.label ?? ''"
+          [required]="false"
+          placeholder="Token label"
+          emptyLabel="No label"
+          textClass="block max-w-lg overflow-hidden text-ellipsis whitespace-nowrap text-2xl font-bold tracking-tight text-gray-900 dark:text-white"
+          [loading]="savingLabel()"
+          (save)="saveLabel($event)" />
+
+        <dl class="grid grid-cols-3 gap-6 mt-4 text-sm">
```
kosmoz, line 75: should maybe still show expiration for legacy tokens

kosmoz (approved):

PR #3234, opened 2026-09-15
fix(backend): allow access token management in SSO sessions

pmig:
…and non access token


internal/middleware/middleware.go
```diff
@@ -294,21 +294,30 @@ func BlockCrossOrganizationAction(handler http.Handler) http.Handler {
...
 
-// CredentialChangeBlockedMessage is the response of BlockCredentialChange, exported for the endpoints that
-// reject only part of their request body and therefore cannot apply the middleware.
-const CredentialChangeBlockedMessage = "your sign-in methods cannot be changed from this session. " +
-	"Request a password reset to receive a link to your email address that lets you change them"
+// The messages are exported for the endpoints that reject only part of their request body and
+// therefore cannot apply the middleware.
+const (
+	OidcSessionBlockedMessage = "your sign-in methods cannot be changed from a session your organization's " +
+		"identity provider signed you in with. Request a password reset to receive a link to your email " +
```
kosmoz: It's not wrong but the first sentence seems a bit clunky and I had to read it three times to understand it. Maybe simplify to something like
```suggestion
	OidcSessionBlockedMessage = "operation not permitted in SSO session. Request a password reset to receive a link to your email " +
```

kosmoz (approved):

PR #3235, opened 2026-09-15
chore: improve access token migration rollover

pmig:


kosmoz (commented):
I think this is an overly complicated solution. Can we not just … not allow to migrate legacy tokens and ask users to create a new one? that would also solve the issue but be way less complicated for us and for the users

kosmoz (approved):

PR #3249, opened 2026-09-16
ci: add scheduled and pr osv scanning

pmig:


kosmoz (commented):
this should be two workflow files 🤔

pmig:
> this should be two workflow files 🤔

ok

kosmoz (approved):

PR #3256, opened 2026-09-16
chore: add support bundle deletion

pmig:


frontend/ui/src/app/support-bundles/support-bundle-display.ts
```diff
@@ -1,6 +1,15 @@
 import {never} from '../../util/exhaust';
+import {ConfirmConfig} from '../components/confirm-dialog/confirm-dialog.component';
 import {SupportBundleStatus} from '../types/support-bundle';
 
+export const supportBundleDeleteConfirm: ConfirmConfig = {
+  message: {
+    message: 'Are you sure you want to delete this support bundle?',
+    alert: {type: 'danger', message: 'All collected resources and comments are deleted permanently.'},
```
kosmoz: Again, I think this should be a warning alert. "danger" should be reserved for more dangerous actions

pmig: ok

kosmoz (approved):

PR #3264, opened 2026-09-17
chore: remove organizationScopeCutoff and always enforce organization id (old)

pmig:


pmig:
https://github.com/distr-sh/distr/pull/3265

PR #3280, opened 2026-09-17
feat(registry): allow public artifacts

pmig:


frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.ts
```diff
@@ -110,7 +116,12 @@ export class ArtifactPullsComponent {
         takeUntilDestroyed(),
         filter((value) => value !== null && value !== '')
       )
-      .subscribe(() => this.filterForm.controls.deploymentTargetId.setValue(''));
+      .subscribe((value) => {
+        this.filterForm.controls.deploymentTargetId.setValue('');
+        if (value === ANONYMOUS_USER) {
+          this.filterForm.controls.customerOrganizationId.setValue('');
+        }
+      });
```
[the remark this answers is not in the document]
pmig: this makes sense

internal/handlers/artifacts.go
```diff
@@ -62,6 +62,14 @@ func ArtifactsRouter(r chiopenapi.Router) {
...
 					}{})).
 					With(option.Response(http.StatusOK, api.ArtifactResponse{}))
+				r.With(middleware.RequireAdmin).
+					Patch("/public", patchArtifactPublicHandler).
+					With(option.Description("Make an artifact publicly pullable or private again")).
+					With(option.Request(struct {
+						ArtifactRequest
+						api.PatchArtifactPublicRequest
+					}{})).
+					With(option.Response(http.StatusOK, api.ArtifactResponse{}))
```
kosmoz, line 72: I think this should be merged with the existing PATCH Artifact request

internal/types/artifact.go
```diff
@@ -29,6 +29,7 @@ type Artifact struct {
 	OrganizationID   uuid.UUID         `db:"organization_id" json:"-"`
 	Name             string            `db:"name" json:"name"`
 	ImageID          *uuid.UUID        `db:"image_id" json:"-"`
+	Public           bool              `db:"public" json:"public"`
```
kosmoz, line 32: Maybe double check but I think the JSON tags here are no longer needed

internal/registry/authn.go
```diff
@@ -0,0 +1,104 @@
...
+
+// AnonymousAccess lets a request that carries no credentials reach the handlers without
+// authentication, so that the authorizer can serve it from a public artifact. Everything else keeps
+// going through authenticated, which answers 401 with the challenge.
+//
+// Only the routes that address an artifact are opened up. The /v2/ ping in particular has to keep
+// its 401: OCI clients record the authentication challenge from that response alone, so a 200 there
+// would leave a client that holds credentials with no challenge to answer and it would never send
+// them, breaking every private pull.
+func AnonymousAccess(
```
kosmoz, line 20: I think keeping the authenticated and non-authenticated middleware stack in different places makes it hard to understand and might introduce bugs in the future. Also, passing the middleware stack intended for authenticated requests to a function called `AnonymousAccess` just looks really odd.

I think it would be nice to add a `middleware.Split()` that accepts a `predicate func(http.Handler) bool` and two `func(http.Handler) http.Handler`. Combined with a `middleware.Chain()` (that I think also doesn't exist yet) this could be used like this:

```go
withMiddlewares(
  // other middlewares...
  middleware.Split(
    isAuthenticated,
    middleware.Chain(
      // authenticated middlewares...
    ),
    middleware.Chain(
      // unauthenticated middlewares...
    )
  )
)
```

what do you think?

kosmoz (commented):
preliminary review (I didn't check everything yet)

internal/registry/authn.go
pmig: done!

frontend/ui/src/app/services/artifacts.service.ts
```diff
@@ -123,6 +124,12 @@ export class ArtifactsService {
       .pipe(tap((it) => this.cache.save(it)));
   }
 
+  public patchPublic(artifactId: string, isPublic: boolean): Observable<ArtifactWithTags> {
+    return this.http
+      .patch<ArtifactWithTags>(`${this.artifactsUrl}/${artifactId}`, {public: isPublic})
+      .pipe(tap((it) => this.cache.save(it)));
+  }
```
kosmoz: I think these should also be merged

internal/registry/authn.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// allowsAnonymous lists the routes that name an artifact, which is what makes the authorizer able
+// to decide anonymous access at all. The catalog belongs to the organization rather than to an
+// artifact and must stay out, as must the ping: OCI clients record the authentication challenge
+// from the /v2/ response alone, so a 200 there would leave a client that holds credentials with no
+// challenge to answer and it would never send them, breaking every private pull.
+func allowsAnonymous(r *http.Request) bool {
+	return isManifest(r) || isBlob(r) || isTags(r) || isReferrers(r)
+}
```
kosmoz, line 52: not sure about the catalog... wouldn't a public catalog also make sense?

frontend/ui/src/app/artifacts/artifact-pulls/artifact-pulls.component.html
```diff
@@ -153,6 +154,8 @@
               <td class="px-4 py-3 font-medium text-gray-900 whitespace-nowrap dark:text-white">
                 @if (pull.userAccountName || pull.userAccountEmail) {
                   {{ pull.userAccountName ?? pull.userAccountEmail }}
+                } @else if (pull.anonymous) {
+                  <app-artifact-anonymous-pull-badge />
```
kosmoz: Personal opinion I guess, but I think the badge looks out of place here. I think just a text "anonymous" with muted grey color or something like that would be more appropriate

internal/registry/authn.go
pmig: As discussed the catalog should not list all public artifacts potentially exposing slugs and images.

internal/types/artifact.go
pmig: no `types.Artifact` is still reference multiple times in artifact api structs

internal/registry/authn.go
```diff
@@ -0,0 +1,82 @@
...
+
+func rateLimitAnonymous(limits env.AnonymousRateLimits) func(http.Handler) http.Handler {
+	return middleware.Chain(
+		limitAnonymous(isManifestOrListing, limits.ManifestsPerMinute, time.Minute),
+		limitAnonymous(isManifestOrListing, limits.ManifestsPerHour, time.Hour),
+		limitAnonymous(isBlob, limits.BlobsPerMinute, time.Minute),
+		limitAnonymous(isBlob, limits.BlobsPerHour, time.Hour),
+	)
+}
```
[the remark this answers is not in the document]
pmig: anonymous users can't push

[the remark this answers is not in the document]
pmig: no

kosmoz (commented):
As discussed, I think there might be a regression with the docker authentication regarding public artifacts. we should double check this

kosmoz (approved):

PR #3281, opened 2026-09-17
feat(backend): agent host

### LINKED ISSUE #3275: Agent API Hostname
Users want to publish the agent API under a different domain name, potentially not fully exposing Distr to the public.

### Discussed in https://github.com/distr-sh/distr/discussions/3163

<div type='discussions-op-text'>

<sup>Originally posted by **will-corrigan** September  4, 2026</sup>
Hey, we currently self host Distr and would like to put it behind Cloudflare Access which uses Okta for human users and a service token for machine users. As far as I understand it both an agent and the UI share the same hostname so it poses a bit of an issue as the agent cannot currently take custom headers to authenticate as a machine user through Cloudflare. I believe this also would stop us installing an agent with the kubectl command as that uses the same hostname.

In other apps I've used they have separate hostnames for things like this (e.g. [datadog](https://docs.datadoghq.com/agent/configuration/network/)), and in Distr you currently have (I think) DISTR_HOST and REGISTRY_HOST to specify this separation for at least these two services. A separate agents one would allow us to specify different policies/allowlists for just agents and allow a better security posture for our situation so it would be quite useful.

Are there any plans for this in the future? If not this, the ability to allow custom headers on agents would also help (albeit be less helpful than the hostname change). 

Thanks for your consideration.</div>

pmig:


website/src/content/docs/docs/self-hosting/configuration.mdx
```diff
@@ -99,10 +98,26 @@ the widget and are not verified; they remain subject to the rate limits of the a
...
+because the UI sits behind a VPN, and point it at a public hostname that routes to the same Distr instance.
+
+It is used for the connect and pre-connect URLs, for the connect command and the pre- and post-connect scripts built
+around them, and for every endpoint in the agent manifest the agents poll afterwards. It takes precedence over an
+organization's custom app domain, so agents keep reaching the public host even for a vendor that serves the UI under
+its own domain. Links, emails and the customer portal are unaffected and keep using `DISTR_HOST` or the
+organization's custom domains.
+
+Existing deployment targets pick the new host up with their next manifest poll. Keep the old host reachable for
+agents until they have, or reconnect them.
```
kosmoz: LLMs started doing this and I don't like it. I think this page should be a simple reference page. long texts like this should be on a separate page that can link to this one.

deploy/charts/distr/values.yaml
```diff
@@ -57,6 +57,10 @@ distr:
   env: &defaultDistrEnv
     - name: DISTR_HOST
       value: http://distr.local
+    # Host agents and connect commands reach Distr at, for an installation whose DISTR_HOST is not
+    # publicly available. Defaults to DISTR_HOST.
```
kosmoz: This sentence is so hard to read. I would honestly just remove this entire comment it, since most selfhosters will not use it and the ones who need it will find it in the docs

internal/env/env.go
```diff
@@ -419,6 +417,11 @@ func HostScheme() URLScheme {
 	return SchemeHTTPS
 }
 
+// AgentHost is the host agents and connect commands reach this instance at, for an installation
+// whose app host is not publicly available. Build a URL on it through
+// customdomains.AgentDomainOrDefault, which keeps the fallback to the app host in one place.
+func AgentHost() *string { return agentHost }
```
kosmoz:
```suggestion
func AgentHost() *string { return agentHost }
```

kosmoz (approved):

PR #3283, opened 2026-09-17
chore: release resources promptly

kojah:
Changes:

- Closing each registry blob immediately after reading its manifest data, and cancel each Compose log context as soon as that deployment's log request completes.
- Release the signal notification contexts when the agent and distr commands finish, so their registered signal handlers don't outlive the operation that owns them.

Testing:

- `go test ./cmd/agent/docker ./cmd/agent/kubernetes ./cmd/distr/cmd ./internal/registry`
- `go run github.com/kojah/gohawk@v0.3.2 ./...`

I'm also working on a static analysis tool called [gohawk](https://gohawk.dev) that I've added to the existing Go lint-and-test job. If you find it helpful, please consider keeping it around! Otherwise, feel free to remove from the PR :)


.github/workflows/build-distr.yaml
```diff
@@ -44,6 +44,8 @@ jobs:
             go.sum
       - name: Run go tests
         run: go test ./...
+      - name: Run gohawk
+        run: go run github.com/kojah/gohawk@v0.3.2 ./...
```
pmig:
```suggestion
```

pmig (changes requested):
Please remove the additional gohawk ci step and feel free to create another pr where we can consider it

kosmoz:
Thanks for the PR, and for being upfront about gohawk in the description.

I don't think this one comes out ahead, though. The registry runs behind `chimiddleware.Recoverer`, so a panic in the handler is recovered and the process keeps serving, which means the blob in `manifest.go` now never gets closed at all. The `defer` was doing the right thing there. The rest of the diff adds `defer stop()` in `main` and the command entry points, where the process exits immediately afterwards, so it has no effect at runtime.

Closing on that basis, but happy to look at future contributions.

On gohawk, we are not looking to add another analyzer next to golangci-lint right now, so please do not spend time on a separate PR for it.

PR #3287, opened 2026-09-18
docs(website): refresh architecture diagrams

thekubernaut:
Replaces the five docs architecture diagrams. They are now built as CSS
components (one per diagram, rendered to transparent 1200x630 PNGs in both
themes), so they can be regenerated and restyled instead of re-exported from
Canva by hand.

## What readers will notice

- Tiles read **Distr** instead of "Distr Hub", matching the v4 rename.
- Vendor group titles read **"Distr SaaS or self-hosted"**.
- In dark mode the whole group-title lockup is white, cloud mark included. It
  used to keep a light-blue cloud next to white text.
- Fixes the typos "Your Applicaton" and "PostrgeSQL".

## Changes beyond a like-for-like refresh

**Assisted Self-Managed** drops the Terraform agent and its Cloud Account row.
The agents overview scopes the Terraform agent to BYOC, so it did not belong in
this scenario. The two remaining sub-groups are re-centred in the customer box.

**BYOC is redesigned.** The old diagram showed a "Terraform Agent" as a peer
agent type alongside Helm and Docker, plus three parallel cloud accounts. What
actually happens is two independent Distr applications:

1. an **infrastructure app** that the ordinary Docker agent deploys onto a
   management VM in the customer's cloud account, where it reconciles their
   infrastructure on a loop;
2. **your application**, managed separately by the Helm agent in the cluster
   that the infrastructure app creates.

Their only link is the reconciler's last step, deploying the Helm agent. The
diagram deliberately carries no Terraform mark: the composition compiles to
Terraform at build time, but the customer never writes HCL, runs `terraform`,
or holds state, so a Terraform logo would advertise the one thing they do not
touch.

**Air-gapped gets a diagram back.** The rewritten page lost its image, leaving
it as the only distribution scenario without a Solution Architecture section.
The new diagram follows the page's own flow: build the offline package in CI,
pull it onto a connected staging host, carry it across the gap, then deploy
offline with Zarf in Kubernetes or `docker load` plus Compose on a VM. No wire
crosses the gap.

## Follow-up needed

The BYOC page and the agents overview still describe the Terraform-agent model
in prose, which now contradicts the BYOC diagram:

- `distribution-scenarios/byoc.mdx` frontmatter description and the paragraph
  starting "After the end customer's cloud account is connected to Distr, the
  Terraform agent provisions the required infrastructure."
- `agents/overview/deployment-agents.mdx`: "**Terraform Agent** _(coming
  soon)_: designed for BYOC deployments", and the sentence above it counting
  "two production-ready agents and one in development".

Deliberately left out of this PR so the wording can be decided separately.

`pnpm format` and `pnpm build` both pass (144 pages, all internal links valid).


website/src/assets/docs/getting-started/what-is-distr/distr-diagram-light.png
kosmoz, line 1: Nit: In the customer boxes, I feel like the blue boxes are not aligned consistently

website/src/assets/docs/getting-started/distribution-scenarios/byoc/byoc-light.png
kosmoz, line 1: terraform agent dropped on purpose?

kosmoz (commented):
Nice change overall!

Not a regression, but I think it's a bit strange that "Distr" and "Distr OCI Registry" are separate boxes throughout. But shouldn't the registry be part of Distr?

website/src/assets/docs/getting-started/distribution-scenarios/byoc/byoc-light.png
thekubernaut: Yes :)

pmig (changes requested):
I would loki to the architecture diagram, which use the same object storage as the registry. additionally i would a png -> webp conversion step at end which will reduce image size

thekubernaut:
Pushed a revision covering all the feedback.

**Loki** (@pmig) is in, sharing the object storage with the registry exactly as `docs/self-hosting` describes it, with its own bucket for log chunks. PostgreSQL, Loki and the object storage now form a dependency column.

**The registry is no longer a sibling of Distr** (@kosmoz). It sits directly under Distr, and the wiring follows the real split: agents connect to Distr, OCI clients connect to the registry. Happy to go further and draw a box around Distr and the registry to make containment explicit, if the stacking isn't enough.

**Tile alignment** (@kosmoz) is fixed. The misalignment was inherited from the Canva originals, where a one-line label makes a 115px tile and a two-line label makes 136px, so aligning tops left the wires off-centre. Tile centres are now aligned per row, so every wire runs through both tiles it joins.

**WebP** (@pmig): lossless, not lossy.

| | PNG | Lossless WebP | Lossy q90 |
|---|---|---|---|
| All ten diagrams | 665 KB | **235 KB** | 355 KB |

Lossless wins outright here because the diagrams are flat colour with sharp text, so there's no reason to trade quality for bytes. Output is byte-identical to the PNGs with transparency intact. Note this also touches `blog/2026-08-19-self-hosted-spectrum.mdx`, which imports the BYOC diagram.

**Terraform agent prose** is rewritten in the same push, since the BYOC diagram contradicted it. The BYOC page now describes the two-application model, and the agents overview says BYOC needs no third agent type because the infrastructure ships as an ordinary Compose application that the Docker agent deploys onto a management VM.

`pnpm format` and `pnpm build` pass (144 pages, links valid).

website/src/assets/docs/getting-started/what-is-distr/distr-diagram-light.png
thekubernaut: Totally fair!

kosmoz (approved):

pmig (approved):

